PR #21, opened 2024-02-01
218364 fix downsizing for multiple new orders

ffaraone:


adobe_vipm/adobe/client.py
```diff
@@ -160,92 +161,119 @@ def create_customer_account(
...
+            new_orders_response.raise_for_status()
+            new_orders_page = new_orders_response.json()
+            for order in new_orders_page["items"]:
+                actual_sku = get_actual_sku(order["lineItems"], sku)
+                if not actual_sku:
+                    continue
+
+                logger.debug(f"Found order to return for sku {actual_sku}: {order['orderId']}")
+
+                item_to_return = next(
```
d3rky: I guess there is an util function `find_first`?

adobe_vipm/adobe/client.py
```diff
@@ -160,92 +161,119 @@ def create_customer_account(
...
+                    filter(
+                        lambda adb_item: adb_item["extLineItemNumber"] == line_number,
+                        order["lineItems"],
+                    ),
+                )
+                order_id = order["externalReferenceId"]
```
d3rky, line 219: Do you think we need these variables? As I can they are used just to construct `external_id`

adobe_vipm/adobe/utils.py
```diff
@@ -0,0 +1,12 @@
+def get_actual_sku(items, sku):
+    item = next(filter(lambda item: item["offerId"].startswith(sku), items), {})
```
d3rky, line 2: `find_first`?

adobe_vipm/adobe/utils.py
```diff
@@ -0,0 +1,12 @@
...
+    return item.get("offerId")
+
+
+def get_item_to_return(items, line_number):
+    return next(
```
d3rky, line 7: `find_first`?

d3rky (changes requested):

adobe_vipm/adobe/utils.py
ffaraone: I will do this refactoring in the next PR. now find_first is located in utils.py of flows package. I don't want to import it in adobe package since my rule is flows use adobe package but not the contrary.

adobe_vipm/adobe/utils.py
ffaraone: Same as before.

adobe_vipm/adobe/client.py
ffaraone: yeah see below.

adobe_vipm/adobe/client.py
ffaraone: Fixed

adobe_vipm/adobe/utils.py
ffaraone: Fixed

adobe_vipm/adobe/utils.py
ffaraone: Fixed

d3rky (approved):

PR #23, opened 2024-02-05
218135 Add opentelemtry and tracing for logger and requests

d3rky:


pyproject.toml
```diff
@@ -15,6 +15,7 @@ requests = "2.31.*"
 django = "4.2.*" # should it be a dependency of the extension ? it is already a dependency of the sdk.
 
 # Remove when devmock is not needed anymore
+azure-monitor-opentelemetry-exporter = "^1.0.0b21"
```
ffaraone, line 18: Why this is a dependency of the extension ?

swo/mpt/extensions/runtime/djapp/conf/default.py
```diff
@@ -132,7 +132,8 @@
             "style": "{",
         },
         "rich": {
```
ffaraone, line 134: IMO when the extension is running NOT in local/dev environment, the rich logging handler must be disabled.
Probably is better to add the OTEL fields to the verbose formatter.

swo/mpt/extensions/runtime/djapp/conf/default.py
```diff
@@ -165,6 +166,13 @@
     },
 }
 
+
+# OpenTelemetry configuration
+SERVICE_NAME = os.getenv("SERVICE_NAME", "Swo.Extensions.AdobeVIPM")
+USE_APPLICATIONINSIGHTS = os.getenv("USE_APPLICATIONINSIGHTS", "True").lower() in ("true", "1", "t")
+APPLICATIONINSIGHTS_CONNECTION_STRING = os.getenv("APPLICATIONINSIGHTS_CONNECTION_STRING", "")
+LOGGING_ATTEMPT_GETTER = os.getenv("LOGGING_ATTEMPT_GETTER", "adobe_vipm.utils.get_attempt_count")
```
ffaraone, line 174: Do you think logging the number attempts is needed from the day 0 ?
When we switch from polling the API to get the events, to push the events to the extensions from the MPT platform, retries, delayed redeliveries etc should be controlled by the MPT platform.

If it is needed from the day 0 let's try to brainstorm if a better approach is possible.

swo/mpt/extensions/runtime/djapp/conf/default.py
```diff
@@ -165,6 +166,13 @@
     },
 }
 
+
+# OpenTelemetry configuration
+SERVICE_NAME = os.getenv("SERVICE_NAME", "Swo.Extensions.AdobeVIPM")
```
ffaraone, line 171: Probably we should think about something like:
Swo.Extensions.<unique_extension_id>
In any case, please change this default to some generic value not related to adobe.

swo/mpt/extensions/runtime/events/utils.py
```diff
@@ -31,5 +44,65 @@ def setup_client():
...
+
+
+def wrap_for_trace(func, event_type):
+    @wraps(func)
+    def wrapper(client, event):
+        tracer = trace.get_tracer(event_type)
+
+        match event_type:
+            case "orders":
+                object_id = event.data["id"]
```
ffaraone, line 88: The Event object already has an id field so I think this logic can be removed.
IMO who is responsible to set the right unique id is the event producer (now the poller, then the pusher).
We should try to let this wrapper as general as we can.

ffaraone (changes requested):

pyproject.toml
d3rky: Agree, auto added

swo/mpt/extensions/runtime/djapp/conf/default.py
d3rky: Well for today we have names of the service for that, but definitely - yes

swo/mpt/extensions/runtime/djapp/conf/default.py
d3rky: Yes, that's an important part since we have a not full cycle of processing like something failed and we reschedule the order somewhere in the middle. For sure in the future, it may be improved by the runtime. But we have what we have now :-)

swo/mpt/extensions/runtime/djapp/conf/default.py
d3rky: added to both, thanks

swo/mpt/extensions/runtime/events/utils.py
d3rky: Cool!

ffaraone (approved):

PR #32, opened 2024-02-23
216758 cancellation out of window

ffaraone:


adobe_vipm/adobe/client.py
```diff
@@ -326,6 +326,7 @@ def create_preview_order(
         reseller_country: str,
         customer_id: str,
         order: dict,
```
d3rky, line 328: WDYT about to have here also an order_id? Now the only property that is required to use by this function is an order_id :-)

adobe_vipm/adobe/errors.py
```diff
@@ -39,9 +40,9 @@ def __repr__(self) -> str:
         return str(self.payload)
 
 
-def wrap_http_error(func: Callable[..., T]):
+def wrap_http_error(func: Callable[Param, RetType]) -> Callable[Param, RetType]:
     @wraps(func)
-    def _wrapper(*args, **kwargs) -> T:
+    def _wrapper(*args: Param.args, **kwargs: Param.kwargs) -> RetType:
```
d3rky, line 45: Decided not to comment it on previous PRs, but you did it by yourself now. Great!

d3rky (approved):

PR #36, opened 2024-02-27
219737 Adjust API and data structures with existing state of MPT API

d3rky:
- enums are lowercase (Change -> change)
 - adjust order/subscription/agreement items structure (items -> lines)
 - use externalId for parameters referencing
 - order --> ordering
 - structured parameter values are passed by json strings

### DIFF
--- a/adobe_vipm/adobe/client.py
+++ b/adobe_vipm/adobe/client.py
@@ -57,23 +57,23 @@ class AdobeClient:
             "externalReferenceId": reseller_id,
             "distributorId": distributor.id,
             "companyProfile": {
-                "companyName": reseller_data["CompanyName"],
-                "preferredLanguage": reseller_data["PreferredLanguage"],
+                "companyName": reseller_data["companyName"],
+                "preferredLanguage": reseller_data["preferredLanguage"],
                 "address": {
-                    "country": reseller_data["Address"]["country"],
-                    "region": reseller_data["Address"]["state"],
-                    "city": reseller_data["Address"]["city"],
-                    "addressLine1": reseller_data["Address"]["addressLine1"],
-                    "addressLine2": reseller_data["Address"]["addressLine2"],
-                    "postalCode": reseller_data["Address"]["postCode"],
-                    "phoneNumber": reseller_data["Contact"]["phone"],
+                    "country": reseller_data["address"]["country"],
+                    "region": reseller_data["address"]["state"],
+                    "city": reseller_data["address"]["city"],
+                    "addressLine1": reseller_data["address"]["addressLine1"],
+                    "addressLine2": reseller_data["address"]["addressLine2"],
+                    "postalCode": reseller_data["address"]["postCode"],
+                    "phoneNumber": reseller_data["contact"]["phoneNumber"],
                 },
                 "contacts": [
                     {
-                        "firstName": reseller_data["Contact"]["firstName"],
-                        "lastName": reseller_data["Contact"]["lastName"],
-                        "email": reseller_data["Contact"]["email"],
-                        "phoneNumber": reseller_data["Contact"]["phone"],
+                        "firstName": reseller_data["contact"]["firstName"],
+                        "lastName": reseller_data["contact"]["lastName"],
+                        "email": reseller_data["contact"]["email"],
+                        "phoneNumber": reseller_data["contact"]["phoneNumber"],
                     }
                 ],
             },
@@ -90,7 +90,7 @@ class AdobeClient:
         created_reseller = response.json()
         adobe_reseller_id = created_reseller["resellerId"]
         logger.info(
-            f"Reseller {reseller_id} - {reseller_data['CompanyName']} "
+            f"Reseller {reseller_id} - {reseller_data['companyName']} "
             "created successfully in distributor account "
             f"{distributor.id} ({region}): {adobe_reseller_id}",
         )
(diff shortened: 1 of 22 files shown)


adobe_vipm/adobe/client.py
```diff
@@ -355,22 +355,22 @@ def create_preview_order(
             "lineItems": [],
         }
 
-        for item in items:
-            product: AdobeProduct = self._config.get_adobe_product(item["productItemId"])
-            quantity = item["quantity"]
-            old_quantity = item["oldQuantity"]
+        for line in lines:
+            product: AdobeProduct = self._config.get_adobe_product(line["item"]["id"])
```
ffaraone, line 359: According to [API docs] (https://softwareone.atlassian.net/wiki/spaces/GPTEAM/pages/5222072917/Order+Object) line["item"]["id"] is the MPT id of the item (i.e. the product item ID) not the MPN that is what we need here to lookup the product from the mapping. IMO using platform id instead of vendor ids make the extension hard to maintain. It should be **item.externalIDs.vendor**  right ?

adobe_vipm/adobe/client.py
```diff
@@ -355,22 +355,22 @@ def create_preview_order(
...
                 # it is placing a new order.
-                # For downsizing items quantity it must send the actual quantity
+                # For downsizing lines quantity it must send the actual quantity
                 # since the previous purchased quantity has been returned back
                 # through one or more RETURN orders.
                 quantity = quantity - old_quantity
             payload["lineItems"].append(
                 {
-                    "extLineItemNumber": item["lineNumber"],
+                    "extLineItemNumber": line["id"],
```
ffaraone, line 373: According to [API docs](https://softwareone.atlassian.net/wiki/spaces/GPTEAM/pages/5222072917/Order+Object) line["id"] has the format ALI-1234-1234-1234-0001 but Adobe expect an integer, we should take just the sequential part of the ID here, right ?

adobe_vipm/flows/fulfillment.py
```diff
@@ -76,20 +77,20 @@ def _create_subscription(mpt_client, seller_country, customer_id, order, item):
...
+                    "externalId": PARAM_SUBSCRIPTION_ID,
                     "value": item["subscriptionId"],
                 }
             ]
         },
-        "items": [
+        "lines": [
             {
-                "lineNumber": item["extLineItemNumber"],
+                "id": item["extLineItemNumber"],
```
ffaraone, line 93: Here item["extLineItemNumber"] is an integer so we should rebuild the MPT line ID.

adobe_vipm/flows/fulfillment.py
```diff
@@ -101,13 +102,13 @@ def _create_subscription(mpt_client, seller_country, customer_id, order, item):
...
-def _update_adobe_subscriptions(seller_country, customer_id, order, items):
+def _update_adobe_subscriptions(seller_country, customer_id, order, lines):
     adobe_client = get_adobe_client()
-    for item in items:
+    for line in lines:
         subcription = get_subscription_by_line_and_item_id(
             order["subscriptions"],
-            item["productItemId"],
-            item["lineNumber"],
+            line["item"]["id"],
```
ffaraone, line 110: this should be the MPN here not the item id.

adobe_vipm/flows/fulfillment.py
```diff
@@ -124,37 +125,37 @@ def _update_adobe_subscriptions(seller_country, customer_id, order, items):
...
     completed_order_ids = []
     pending_order_ids = []
-    for item in items:
+    for line in lines:
         orders_4_item = adobe_client.search_new_and_returned_orders_by_sku_line_number(
             seller_country,
             customer_id,
-            item["productItemId"],
-            item["lineNumber"],
+            line["item"]["id"],
```
ffaraone, line 146: same here MPN (**item.externalIDs.vendor**)

adobe_vipm/flows/fulfillment.py
```diff
@@ -350,9 +351,11 @@ def _fulfill_termination_order(mpt_client, seller_country, order):
 
 def fulfill_order(client, order):
     logger.info(f'Start processing {order["type"]} order {order["id"]}')
-    seller_id = order["agreement"]["seller"]["id"]
-    seller = get_seller(client, seller_id)
-    seller_country = seller["address"]["country"]
+    if "parameters" in order:
+        order["parameters"] = unpack_structured_parameters(order["parameters"])
+    agreement = get_agreement(client, order["agreement"]["id"])
```
ffaraone, line 356: Why you restored the get of the agreement ? In theory the poller send a select=+agreement and the API seems to understand it

adobe_vipm/flows/mpt.py
```diff
@@ -1,6 +1,90 @@
...
+            )
+
+            return entity
+
+        return _wrapper
+
+    return _decorator
+
+
+@unpack_decorator("parameters")
```
ffaraone, line 80: Since you have to add the logic to pack inside the function I would prefer also to call the unpack inside the function instead of using the decorator because IMO is more clear. Anyway up to you, it's a temporary solution.

adobe_vipm/flows/utils.py
```diff
@@ -17,10 +17,10 @@
 from adobe_vipm.utils import find_first
 
 
-def get_parameter(object, parameter_phase, parameter_name):
+def get_parameter(order, parameter_phase, param_external_id):
```
ffaraone, line 20: we can left here object instead of order since the function applies both to orders and subscriptions, right ?

ffaraone (changes requested):

ffaraone (changes requested):
Enormous job, thank you!

adobe_vipm/adobe/client.py
d3rky: Yup, you are correct here, my friend. Good catch! :-)

adobe_vipm/adobe/client.py
d3rky: Wow, it means I have to get it from the product definition :-(

adobe_vipm/adobe/client.py
d3rky: added `to_adobe_line_id`

adobe_vipm/flows/mpt.py
d3rky: yup, in order not to change functions I added decorators to pack/unpack. Easy to remove ;-)

adobe_vipm/flows/utils.py
d3rky: Agree :-)

ffaraone (approved):
🥇

PR #39, opened 2024-03-01
220005 Adjust complex parameters value structure

d3rky:


adobe_vipm/flows/fulfillment.py
```diff
@@ -189,7 +189,7 @@ def _place_new_order(mpt_client, seller_country, customer_id, order):
     adobe_order_id = adobe_order["orderId"]
     order = set_adobe_order_id(order, adobe_order_id)
     logger.debug(f'Updating the order {order["id"]} to save order id into vendor external id')
-    order = update_order(mpt_client, order["id"], externalIds=order["externalIds"])
+    update_order(mpt_client, order["id"], externalIds=order["externalIds"])
```
d3rky, line 192: PUT returns not full representation of the order. Since we have everything in this object and update all the fields during processing no need to return it or make extra call to the API

ffaraone (approved):

PR #42, opened 2024-03-05
219743 add jwt auth to draft validation endpoint

ffaraone:


swo/mpt/extensions/runtime/djapp/conf/default.py
```diff
@@ -173,9 +167,16 @@
...
+# Proxy settings
+USE_X_FORWARDED_HOST = True
+SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
+
+# MPT settings
+
 MPT_API_BASE_URL = os.getenv("MPT_API_BASE_URL", "http://localhost:8000")
 MPT_API_TOKEN = os.getenv("MPT_API_TOKEN", "change-me!")
-PRODUCT_ID = os.getenv("PRODUCT_ID", "PRD-1111-1111-1111")
+MPT_PRODUCT_ID = os.getenv("MPT_PRODUCT_ID", "PRD-1111-1111-1111")
```
d3rky, line 178: We should notify Sylwester about this changes

swo/mpt/extensions/core/security.py
```diff
@@ -0,0 +1,38 @@
...
+        self.secret_callback = secret_callback
+        super().__init__()
+
+    def authenticate(self, request: HttpRequest, token: str) -> Any | None:
+        audience = request.get_host()
+        try:
+            claims = jwt.decode(
+                token,
+                audience=audience,
+                options={"verify_signature": False},
```
d3rky, line 27: Are you sure here?

d3rky (changes requested):

swo/mpt/extensions/core/security.py
ffaraone: Yes, the idea is the following:
* Decode the token to get the claims (no signature verification)
* Call the jwt_secret_callback with claims, claims contain the webhook id so the callback can return the corresponding secret
* Decode with the secret (it will verify signature, expiration, nbf and audience)

swo/mpt/extensions/runtime/djapp/conf/default.py
ffaraone: Yeah you are right, I modified it in the TDRs but I forgot to mention it in the changes notification email.
I will create a teams group with Sylvester you and me to stay in sync until the first deployment succeed.

PR #43, opened 2024-03-05
219735 Fix docker image and docker commands for bash, app, tests, for…

d3rky:


dev.Dockerfile
```diff
@@ -3,11 +3,12 @@ ENV PYTHONUNBUFFERED=1 POETRY_VERSION=1.7.0
...
 
-WORKDIR /install_temp
-COPY pyproject.toml poetry.* /install_temp/
+WORKDIR /extension
 
-RUN poetry update && poetry install --with dev,devmock,runtime,sdk
+ADD . /extension
 
-ADD ./adobe_vipm /extension
+RUN poetry update && poetry install --with dev,devmock,runtime,sdk && poetry install --only-root
```
ffaraone, line 10: The last `poetry install --only-root` in theory is not needed because the previous `poetry install --with dev,devmock,runtime,sdk` include the root installation unless you call it with --no-root.

prod.Dockerfile
```diff
@@ -3,11 +3,12 @@ ENV PYTHONUNBUFFERED=1 POETRY_VERSION=1.7.0
...
 
-WORKDIR /install_temp
-COPY pyproject.toml poetry.* /install_temp/
+WORKDIR /extension
 
-RUN poetry update && poetry install --with runtime,sdk,devmock
+ADD . /extension
 
-ADD ./adobe_vipm /extension
+RUN poetry update && poetry install --with dev,devmock,runtime,sdk && poetry install --only-root
```
ffaraone, line 10: Same as in dev.Dockerfile

ffaraone (approved):
I would prefer to have an entrypoint.sh script and have bash or swext as part of the command.

Imagine you want to enter the bash (without using a docker compose)

The most natural command line is:

```
$ docker run -it <image-name> bash
```

Instead of:

```
$ docker run -it  --entrypoint bash <image-name>
```

prod.Dockerfile
d3rky: Done

dev.Dockerfile
d3rky: Done

d3rky:
> I would prefer to have an entrypoint.sh script and have bash or swext as part of the command.
> 
> Imagine you want to enter the bash (without using a docker compose)
> 
> The most natural command line is:
> 
> ```
> $ docker run -it <image-name> bash
> ```
> 
> Instead of:
> 
> ```
> $ docker run -it  --entrypoint bash <image-name>
> ```

Thanks, my friend! Done

PR #49, opened 2024-03-07
220760 Remove externalIds query, now it is a port of the Item reference

d3rky:


devmock/mockgendata.py
```diff
@@ -181,7 +179,7 @@ def gen_address(fake):
     return {
         "addressLine1": fake.street_address(),
         "addressLine2": fake.secondary_address(),
-        "postalCode": postcode,
+        "postCode": postcode,
```
d3rky, line 182: Woops, made mistake on previous PR :-(

ffaraone (approved):

PR #54, opened 2024-03-13
220010 implement authentication using login/paswd

d3rky:


swo/mpt/client/base.py
```diff
@@ -14,12 +15,34 @@ def __init__(self, base_url, api_token):
...
 
     def request(self, method, url, *args, **kwargs):
+        if self._api_token is None:
+            self.authorize()
+
         url = self.join_url(url)
+        response = super().request(method, url, *args, **kwargs)
+
+        if response.status_code == 401:
+            self.authorize()
```
ffaraone, line 44: Probably here we should return response if status_code != 401 right ? Otherwise we make always two calls!

ffaraone (changes requested):

ffaraone (approved):

PR #63, opened 2024-03-21
221682 mock pricelist listing apis

ffaraone:


adobe_vipm/flows/mpt.py
```diff
@@ -15,7 +15,7 @@ def _has_more_pages(page):
 @wrap_http_error
 def get_agreement(mpt_client, agreement_id):
     response = mpt_client.get(
-        f"/commerce/agreements/{agreement_id}?select=seller,buyer"
+        f"/commerce/agreements/{agreement_id}?select=seller,buyer,listing"
```
d3rky, line 18: Is it going to fail existing solution, while listings are not there?

d3rky (changes requested):

adobe_vipm/flows/mpt.py
ffaraone: I will revert for the moment

d3rky (approved):

PR #64, opened 2024-03-22
221774 update subscription purchase unit price

ffaraone:


adobe_vipm/flows/mpt.py
```diff
@@ -113,3 +113,14 @@ def get_product_items_by_skus(mpt_client, product_id, skus):
         offset += limit
 
     return items
+
+
+@wrap_http_error
+def get_pricelist_item_by_product_item(mpt_client, pricelist_id, product_item_id):
+    url = f"/price-lists/{pricelist_id}/price-items?eq(item.id,{product_item_id})"
```
d3rky, line 120: As I remember we should search by full sku (that we store in a special parameter) not by item id, because we are going to store non sellable items in pricelist with proper discount levels

d3rky (changes requested):

adobe_vipm/flows/mpt.py
d3rky: Let's chat and have some document about it first :-)

ffaraone: In pricelist item there is no SKU, the full SKU is in items that are not available for selling. So to update the price we first lookup such items by full SKU then we lookup the pricelist item using the item we get in the previous lookup.

PR #75, opened 2024-04-17
222816 add migration commands

ffaraone:


adobe_vipm/flows/migration.py
```diff
@@ -0,0 +1,163 @@
...
+    get_transfers_to_process,
+)
+
+RECOVERABLE_TRANSFER_ERRORS = (
+    "RETURNABLE_PURCHASE",
+    "IN_WINDOW_NO_RENEWAL",
+    "IN_WINDOW_PARTIAL_RENEWAL",
+    "EXTENDED_TERM_3YC",
+)
+RUNNING_MAX_RETRIES = 15
```
d3rky, line 25: Agreed offline to make it configurable

d3rky (changes requested):

adobe_vipm/flows/migration.py
ffaraone: Fixed

d3rky (approved):

PR #86, opened 2024-04-25
223849 Squash configuration variables to one for different products

d3rky:


swo/mpt/extensions/runtime/initializer.py
```diff
@@ -7,10 +8,18 @@
 from swo.mpt.extensions.runtime.utils import get_extension_app_config_name
 
 
+JSON_EXT_VARIABLES = {"EXT_QUERYING_TEMPLATES_IDS", "EXT_COMPLETED_TEMPLATES_IDS", "EXT_WEBHOOKS_SECRETS"}
```
ffaraone: This shouldn't be here but anyway we can fix later.

ffaraone (approved):

PR #119, opened 2024-05-13
224637 check 3yc commitment and recommitment requests, resubmit decli…

ffaraone:
…ned, submit recommitment requests


adobe_vipm/management/commands/process_3yc.py
```diff
@@ -0,0 +1,33 @@
...
+class Command(BaseCommand):
+    help = "Process 3-year commitment and recommitment requests"
+
+    def success(self, message):
+        self.stdout.write(self.style.SUCCESS(message), ending="\n")
+
+    def info(self, message):
+        self.stdout.write(message, ending="\n")
+
+    def handle(self, *args, **options):
```
d3rky, line 20: My friend, I think it is better here to have a date as an input parameter with default to `today()`. It always simplify maintenance if we compare with hardcoded `today()` in the code

d3rky (approved):

d3rky (approved):

PR #152, opened 2024-06-05
226126 add support for onetime items

ffaraone:


adobe_vipm/flows/fulfillment/change.py
```diff
@@ -288,31 +289,36 @@ def fulfill_change_order(mpt_client, order):
...
         updated_lines.append(order_line)
-        order_subscription = get_subscription_by_line_and_item_id(
-            order["subscriptions"],
-            order_line["item"]["id"],
-            order_line["id"],
-        )
-        if not order_subscription:
-            add_subscription(
-                mpt_client, adobe_client, customer_id, order, ITEM_TYPE_ORDER_LINE, item
+        if item["offerId"][:10] not in one_time_skus:
```
d3rky, line 303: I would recommend here to add a separate function to extract basic sku. Similar code is already exists in 5 places and i think it is time to extract it and avoid +1 place :-) Agree?

CleanShot 2024-06-05 at 11 18 57

adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -387,7 +388,7 @@ def switch_order_to_completed(mpt_client, order, template_name):
     logger.info(f'Order {order["id"]} has been completed successfully')
 
 
-def add_subscription(mpt_client, adobe_client, customer_id, order, item_type, item):
+def     add_subscription(mpt_client, adobe_client, customer_id, order, item_type, item):
```
d3rky, line 391: I think we don't need these changes

d3rky (changes requested):

d3rky (approved):

PR #208, opened 2024-08-01
MPT-120 lookup prices in AirTable during draft validation

ffaraone:


d3rky:
Wow, how simple it becomes!

d3rky (approved):

PR #214, opened 2024-08-07
MPT-425 Introduce command argument to run either API or Consumer pods

d3rky:


ffaraone (approved):
IMO, the "command" name/metavar is a bit redundant here, probably a better name can be found, but due to the urgency of the task definitely not today.

PR #215, opened 2024-08-07
MPT-905 use commitment data on transfer draft validation for prices

ffaraone:


ffaraone:
Recently Adobe made a change to the VIP to VIPM transfer preview API to include into the response commitment data for 3yc customers. Due to that we can use such data to lookup historical price items according to 3yc policies during draft validation of migration purchase order.

d3rky (approved):

PR #229, opened 2024-08-29
MPT-2123 purchase fulfillment now uses pipeline

ffaraone:


adobe_vipm/flows/fulfillment/purchase.py
```diff
@@ -32,302 +31,243 @@
...
+    VIPM API.
+    """
+
+    def __call__(self, client, context, next_step):
+        adobe_client = get_adobe_client()
+        context.adobe_customer = adobe_client.get_customer(
+            context.authorization_id,
+            context.adobe_customer_id,
+        )
+        next_step(client, context)
```
d3rky, line 76: BTW, I was thinking about it. Do you think that using a return statement is better to control the flow? In this case, the Pipeline code itself should switch to the next step. What is needed from the step code is to return updated context. Otherwise we should be strict here to always call `next_step`

d3rky (approved):

adobe_vipm/flows/fulfillment/purchase.py
ffaraone: next_step will be called from the current step only if the current_step completed successfully. For example the SubmitNewOrder step only go to the next one when Adobe reports the order has been processed. Otherwise it stop the processing. IMO returning or not returning the context to make the pipeline understand if it has to continue or stop is more or less the same, probably a bit less clear.

PR #242, opened 2024-09-23
MPT-2980 check incompatible items retrieving a preview order before p…

ffaraone:
…rocessing

### DIFF
--- a/adobe_vipm/flows/fulfillment/shared.py
+++ b/adobe_vipm/flows/fulfillment/shared.py
@@ -18,6 +18,7 @@ from adobe_vipm.adobe.constants import (
     STATUS_PROCESSED,
     UNRECOVERABLE_ORDER_STATUSES,
 )
+from adobe_vipm.adobe.errors import AdobeError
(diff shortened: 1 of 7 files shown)


adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -710,6 +711,29 @@ def __call__(self, client, context, next_step):
...
 
+class GetPreviewOrder(Step):
+    """
+    Retrieve a preview order for the upsize lines. Switch the order to failed
+    in case of exceptions.
+    """
+
+    def __call__(self, client, context, next_step):
+        adobe_client = get_adobe_client()
+        if context.upsize_lines and not context.adobe_new_order_id:
```
d3rky, line 722: Can you please put more comments on why we have this here? It is not clear, especially because your are using this  step only for purchase pipe

d3rky (changes requested):

d3rky (approved):

d3rky (approved):

PR #252, opened 2024-10-17
MPT-3758, MPT-3759 Add new logic to the migration of already transfer…

ruben-sebrango:
Migrating an already transferred Adobe account with an expired subscription will require additional validations to avoid manual processes.

### DIFF
--- a/adobe_vipm/flows/validation/transfer.py
+++ b/adobe_vipm/flows/validation/transfer.py
@@ -22,15 +22,19 @@ from adobe_vipm.flows.constants import (
     ERR_ADOBE_MEMBERSHIP_ID_ITEM,
     ERR_ADOBE_MEMBERSHIP_NOT_FOUND,
     ERR_ADOBE_UNEXPECTED_ERROR,
+    ERR_UPDATING_TRANSFER_ITEMS,
     PARAM_MEMBERSHIP_ID,
 )
 from adobe_vipm.flows.mpt import get_product_items_by_skus
 from adobe_vipm.flows.utils import (
+    are_all_transferring_items_expired,
     get_adobe_membership_id,
+    get_one_time_skus,
     get_order_line_by_sku,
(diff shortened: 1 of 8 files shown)


adobe_vipm/flows/validation/transfer.py
```diff
@@ -282,4 +329,12 @@ def validate_transfer(mpt_client, adobe_client, order):
         subscription["offerId"] = correct_sku or subscription["offerId"]
     customer = adobe_client.get_customer(authorization_id, transfer.customer_id)
     commitment = get_3yc_commitment(customer)
-    return add_lines_to_order(mpt_client, order, subscriptions, commitment, "currentQuantity")
+
+    one_time_skus = get_one_time_skus(mpt_client, order)
```
d3rky, line 333: Talked privately - there is an issue that now all one-time items are removed from transfer. It should be only in the case if all subscriptios are expird

d3rky (changes requested):

adobe_vipm/flows/validation/transfer.py
```diff
@@ -69,37 +73,89 @@ def get_prices(order, commitment, adobe_skus):
...
+
+    adobe_items_map = {
+        get_partial_sku(adobe_item["offerId"]): adobe_item[quantity_field]
+        for adobe_item in adobe_items
+    }
+    return order_line_map != adobe_items_map
 
-def add_lines_to_order(mpt_client, order, adobe_object, commitment, quantity_field):
+
+def add_lines_to_order(mpt_client,
```
d3rky, line 101: Please, run black formatter, as I remember we have another rule to format function definition here

d3rky (changes requested):
Please, fix the formatting

d3rky (approved):

d3rky (approved):

PR #254, opened 2024-10-22
Client can downsize items quantity lower than the required commitment for 3y

ruben-sebrango:
Add new logic to validate if the client downsizes items quantity is lower than the minimum quantity required for 3YC. This new validation has been included for change and termination orders


adobe_vipm/flows/helpers.py
```diff
@@ -130,3 +144,123 @@ def __call__(self, client, context, next_step):
...
+    is not allowed below the minimum commitment of licenses and consumables.
+    """
+
+    def __init__(self, is_validation=False):
+        self.is_validation = is_validation
+
+    def __call__(self, client, context, next_step):
+
+        # Check if the Adobe customer has a 3YC commitment
+        commitment = get_3yc_commitment(context.adobe_customer)
```
d3rky, line 161: Just curios, why you don't want to check in order parameter side?

d3rky (approved):

adobe_vipm/flows/helpers.py
```diff
@@ -130,3 +144,123 @@ def __call__(self, client, context, next_step):
...
+        if commitment and context.downsize_lines:
```
d3rky, line 162: Looks like we need to check status here also

d3rky (changes requested):

d3rky (approved):

PR #265, opened 2024-10-29
MPT-4085 Fix SonarCloud issues

ruben-sebrango:


adobe_vipm/flows/helpers.py
```diff
@@ -159,29 +159,19 @@ def __init__(self, is_validation=False):
...
+        # Get the 3YC commitment if it is enabled
+        commitment = self.get_3yc_commitment_enabled(context.adobe_customer)
 
-        if (
-            commitment
-            and commitment["status"] in (STATUS_3YC_COMMITTED, STATUS_3YC_ACTIVE)
-            and date.today() <= date.fromisoformat(commitment["endDate"])
-            and context.downsize_lines
-        ):
+        if commitment and context.downsize_lines:
```
d3rky, line 165: Please, can you put more details on this change - you remove 2 from 4 if conditions

d3rky (changes requested):

adobe_vipm/flows/helpers.py
ruben-sebrango: I have created a new method "get_3yc_commitment_enabled" to return the commitment only if it is enabled. These two conditions have been moved to a new method to reduce the method's complexity

d3rky (approved):

PR #288, opened 2024-12-12
MPT-5134 Adobe VIPM: Global Customers - Create Airtable models

ruben-sebrango:


adobe_vipm/flows/airtable.py
```diff
@@ -429,3 +511,108 @@ def get_prices_for_3yc_skus(product_id, currency, start_date, skus):
         if item.sku not in prices:
             prices[item.sku] = item.unit_pp
     return prices
+
+
+def create_global_customer_agreement_deployment(product_id, agreement_deployment):
```
robcsegal, line 516: Should we update these create functions to be batch saves with a dictionary list? What do you think?

robcsegal (approved):

adobe_vipm/flows/airtable.py
ruben-sebrango: I agree with you! I have updated the method. Thank you! :-)

adobe_vipm/flows/airtable.py
```diff
@@ -429,3 +511,113 @@ def get_prices_for_3yc_skus(product_id, currency, start_date, skus):
...
+        AirTableBaseInfo.for_migrations(product_id)
+    )
+    return GlobalCustomerAgreementDeployment.all(
+        formula=AND(
+            EQUAL(FIELD("main_agreement_id"), STR_VALUE(main_agreement_id)),
+        ),
+    )
+
+
+def get_global_customer_agreement_deployments_to_check(product_id):
```
d3rky, line 605: Probably let's intriduce a shortcut for `global customer agreement` --> gca? or just global_agreement?

PR #290, opened 2024-12-17
[MPT-5135] Bulk migration tool adds new global customer if account transferred is a global customer

robcsegal:
If the account transferred is a Global customer, a new logic will be implemented to add the information for this transferred account to the new Global Customer main Agreements Airtable in status “pending”.

Create the agreement deployment without subscriptions (item). you need to add the new line in the gc_main_agreement if the customer has "globalSalesEnabled".


adobe_vipm/flows/migration.py
```diff
@@ -343,6 +348,36 @@ def check_running_transfers_for_product(product_id):
...
+                    "membership_id": transfer.membership_id,
+                    "transfer_id": transfer.transfer_id,
+                    "customer_id": transfer.customer_id,
+                    "status": STATUS_GC_PENDING,
+                    "authorization_uk": transfer.authorization_uk,
+                }
+                try:
+                    create_gc_main_agreement(product_id, gc_main_agreement_data)
+                except AirTableHttpError as e:
+                    transfer.status = "failed"
```
ruben-sebrango, line 368: If there is any error managing the Global Customer feature, the migration should continue, and the issue should be reported to the teams. We cannot change the transfer status to fail because the Adobe account has been migrated.

ruben-sebrango (changes requested):

ruben-sebrango (approved):

PR #305, opened 2025-01-15
#MPT-4430 sync agreement issues are not reported to the adobe teams channel

albertsola:


adobe_vipm/flows/utils.py
```diff
@@ -641,6 +641,19 @@ def notify_unhandled_exception_in_teams(process, order_id, traceback):
     )
 
 
+@functools.cache
+def notify_agreement_unhandled_exception_in_teams(process, agreement_id, traceback):
+    """
+    Notify that an agreement has been unhandled exception
+    """
+    send_exception(
+        f"Agreement {process} unhandled exception!",
```
d3rky, line 650: I would remove process name here

d3rky (changes requested):

d3rky (changes requested):

PR #326, opened 2025-02-03
[MPT-4225] Cover extensions runtime with tests

robcsegal:
This part of swo-adobe-vipm-extension isn’t covered by tests. https://github.com/softwareone-platform/swo-adobe-vipm-extension/tree/main/swo/mpt

https://softwareone.atlassian.net/browse/MPT-4225

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,10 +1,15 @@
 import copy
+import json
+import signal
 from collections import defaultdict
 from datetime import UTC, date, datetime, timedelta
 
 import jwt
 import pytest
 import responses
+from django.conf import settings
+from rich.highlighter import ReprHighlighter as _ReprHighlighter
+from swo.mpt.extensions.core.events.dataclasses import Event
 from swo.mpt.extensions.runtime.djapp.conf import get_for_product
 
 from adobe_vipm.adobe.client import AdobeClient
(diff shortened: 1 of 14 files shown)


ruben-sebrango (approved):

tests/conftest.py
```diff
@@ -1511,3 +1514,174 @@ def _mocked_cache(cache=None):
...
+    ]
+
+
+@pytest.fixture()
+def mock_runtime_master():
+    color = True
+    debug = False
+    reload = True
+    component = "all"
+    return Master(
```
d3rky, line 1552: I would suggest just including all values in the dictionary definition.It makes not a lot of sense to have  variables here

tests/core/test_core_registry.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def test_registry_listener():
+    registry = EventsRegistry()
+
+    @registry.listener("orders")
+    def test_listener(client, event):
+        pass
+
+    assert registry.get_listener("orders") == test_listener
```
d3rky, line 17: The recommended way it to test one time in one test. In ideal world it means to have only one `assert` in the test. That's ok to have multiple, but let's test only one function per test

tests/runtime/events/test_events_dispatcher.py
```diff
@@ -0,0 +1,77 @@
...
+    assert isinstance(dispatcher.executor, ThreadPoolExecutor)
+    assert isinstance(dispatcher.queue, deque)
+    assert dispatcher.futures == {}
+    assert isinstance(dispatcher.running_event, threading.Event)
+    assert isinstance(dispatcher.processor, threading.Thread)
+
+
+def test_dispatcher_running():
+    is_success = True
+    try:
```
d3rky, line 23: You really don't need to have hare try/except block. If test raise exception it is going to be failed. In this case you really loosing information from exception

tests/runtime/events/test_events_producers.py
```diff
@@ -0,0 +1,111 @@
...
+
+def test_event_producer_get_processing_orders(
+    mpt_client,
+    mock_wrap_event,
+    requests_mocker,
+    order_factory,
+):
+    order = order_factory()
+
+    products = ",".join(settings.MPT_PRODUCTS_IDS)
```
d3rky, line 32: It looks like this mocker definition is a good candidate for a fixture. At least you have it multiple times here

tests/runtime/events/test_events_utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+def test_response_hook(mocker):
+    mock_span = mocker.MagicMock()
+    mock_request = HttpRequest()
+    mock_response = HttpResponseBase()
+    mock_response.ok = True
+    mock_request.META["x-request-id"] = "1234"
+    mock_request.META["x-correlation-id"] = "5678"
+    mock_request._body = {"data": "test body"}
+    _response_hook(mock_span, mock_request, mock_response)
```
d3rky, line 45: Better to avoid testing of the not "public" functions. Started from `_` or `__`

tests/runtime/events/test_events_utils.py
```diff
@@ -0,0 +1,46 @@
...
+        assert event == mock_wrap_event
+
+    event_type = "orders"
+    settings.USE_APPLICATIONINSIGHTS = appinsights
+    wrapped_func = wrap_for_trace(func, event_type)
+    wrapped_func(client=mpt_client, event=mock_wrap_event)
+    assert event_type == "orders"
+
+
+def test_instrument_logging():
```
d3rky, line 26: Didn't get what are you going to test here. Probably better just to comment to exclude from coverage

tests/runtime/test_runtime_initializer.py
```diff
@@ -0,0 +1,47 @@
...
+
+    monkeypatch.setenv("EXT_WEBHOOKS_SECRETS", webhook_secret)
+    monkeypatch.setenv("EXT_AIRTABLE_BASES", airtable_base)
+    monkeypatch.setenv("EXT_AIRTABLE_PRICING_BASES", airtable_pricing_base)
+    monkeypatch.setenv("EXT_PRODUCT_SEGMENT", product_segment)
+    monkeypatch.setenv("EXT_EMAIL_NOTIFICATIONS_SENDER", email_sender)
+
+    extension_variables = get_extension_variables()
+
+    assert extension_variables is not None
```
d3rky, line 22: I would suggest to have here one `assert` comparing whole dictionary here

tests/runtime/test_runtime_logging.py
```diff
@@ -0,0 +1,31 @@
...
+    repr_highlighter = ReprHighlighter()
+    assert repr_highlighter.accounts_prefixes == mock_logging_account_prefixes
+    assert repr_highlighter.catalog_prefixes == mock_logging_catalog_prefixes
+    assert repr_highlighter.commerce_prefixes == mock_logging_commerce_prefixes
+    assert repr_highlighter.aux_prefixes == mock_logging_aux_prefixes
+    assert repr_highlighter.all_prefixes == mock_logging_all_prefixes
+    assert repr_highlighter.highlights == mock_highlights
+
+
+def test_rich_handler():
```
d3rky, line 29: I'm not sure, that it is needed to be covered

tests/runtime/test_runtime_master.py
```diff
@@ -0,0 +1,33 @@
...
+        mock_runtime_master.setup_signals_handler()
+    except Exception:
+        is_success = False
+    assert is_success
+
+@override_settings(
+    MPT_PRODUCTS_IDS="PRD-1111-1111",
+    LOGGING={"loggers": {"swo.mpt": {}}}
+)
+def test_master_start_stop_signals_handler(mock_runtime_master):
```
d3rky, line 22: Same here, you are basically checks nothing here...

tests/runtime/test_runtime_utils.py
```diff
@@ -0,0 +1,50 @@
...
+    get_events_registry,
+    get_extension,
+    get_extension_app_config_name,
+    get_extension_appconfig,
+    gradient,
+    show_banner,
+)
+
+
+def test_gradient(mock_gradient_result):
```
d3rky, line 13: I'm not sure that it is required to be covered also

d3rky (changes requested):

tests/core/test_core_registry.py
robcsegal: assertions updated and additional tests added where needed

tests/conftest.py
robcsegal: updated to dict

tests/runtime/events/test_events_dispatcher.py
robcsegal: removed

tests/runtime/events/test_events_utils.py
robcsegal: removed

tests/runtime/test_runtime_initializer.py
robcsegal: assertion updated to use dict

tests/runtime/test_runtime_master.py
robcsegal: removed

tests/runtime/events/test_events_producers.py
robcsegal: producer tests updated

tests/runtime/events/test_events_utils.py
robcsegal: test removed

tests/runtime/test_runtime_utils.py
robcsegal: test removed

PR #331, opened 2025-02-13
#MPT-6939 Remove hardcoded SKU from config

albertsola:


ruben-sebrango (commented):
@albertsola I would like to understand how the implemented cache works. The cache should work for all threads, not just for thread execution. That is, the first execution should call Airtable and add the SKUs to the cache, the rest of the orders created on the same day should get the SKUs from the cache.
Could you confirm if the cache works only for the current thread?

adobe_vipm/airtable/models.py
```diff
@@ -20,6 +20,10 @@
 
 from adobe_vipm.utils import find_first
 
+from ..adobe.errors import AdobeProductNotFoundError
```
d3rky, line 23: Please, use absolute imports here :-)

d3rky (changes requested):

adobe_vipm/airtable/models.py
```diff
@@ -659,3 +670,60 @@ def get_agreement_deployment_view_link(product_id):
         return f"https://airtable.com/{base_id}/{table_id}/{view_id}/{record_id}"
     except HTTPError:
         pass
+
+
+@cache
+def get_sku_adobe_mapping_model(
+    base_info: AirTableBaseInfo,
+) -> AdobeProductMappingProtocol:
```
d3rky, line 678: The thing is, that if you are introducing it you should refactor all the tables that we have... But do you really want it? Especially keeping in mind, that for extension code typing is optional

d3rky (changes requested):

adobe_vipm/airtable/models.py
albertsola: With AdobeProductMappingProtocol I replaced AdobeProduct dataclass but I will delete this for now.

d3rky (approved):

PR #334, opened 2025-02-14
MPT-7003 Downsizes within the last two weeks before renewal look for …

ruben-sebrango:
…returnable orders when they should just set renewal quantity

### DIFF
--- a/adobe_vipm/flows/fulfillment/change.py
+++ b/adobe_vipm/flows/fulfillment/change.py
@@ -35,6 +35,7 @@ from adobe_vipm.flows.pipeline import Pipeline, Step
 from adobe_vipm.flows.utils import (
     get_adobe_subscription_id,
     get_subscription_by_line_and_item_id,
+    is_within_last_two_weeks,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


adobe_vipm/flows/fulfillment/change.py
```diff
@@ -54,6 +55,20 @@ class GetReturnableOrders(Step):
     def __call__(self, client, context, next_step):
         adobe_client = get_adobe_client()
         returnable_orders_count = 0
+
+        last_two_weeks = (
+            datetime.fromisoformat(context.adobe_customer["cotermDate"])
+            - timedelta(days=13)
```
d3rky, line 61: Please, extract it to the constant :-)

adobe_vipm/flows/fulfillment/change.py
```diff
@@ -54,6 +55,20 @@ class GetReturnableOrders(Step):
...
+        ).date()
+
+        if date.today() >= last_two_weeks:
+            logger.info(
+                "Downsize occurs in the last two weeks before the anniversary date. "
```
d3rky, line 66: Since this is going to be in logs, can you please put more details here. Like Returnable orders are not going to be submitted,

d3rky (changes requested):

d3rky (approved):

PR #346, opened 2025-03-07
MPT-7723 Use OTEL_SERVICE_NAME environment variable to setup the service name

d3rky:


robcsegal (approved):

tests/runtime/events/test_events_producers.py
```diff
@@ -6,38 +6,6 @@
 )
 
 
-def test_event_producer_produce_events(
```
d3rky, line 9: Removing this test, just because it failing without any issue/stacktrace and tests keep been green

PR #352, opened 2025-03-20
Internal server error for get customers should not fail order

robcsegal:
Order shouldn’t be failed in case of Adobe API instability. 500 for GET /customers/{id} led to order failure.
error message: 
{"code":"1124","message":"Internal Server Error","additionalDetails":[]}

https://softwareone.atlassian.net/browse/MPT-7503


adobe_vipm/flows/helpers.py
```diff
@@ -163,13 +169,21 @@ def __call__(self, client, context, next_step):
...
-        next_step(client, context)
+            try:
+                context.adobe_customer = adobe_client.get_customer(
+                    context.authorization_id,
+                    context.adobe_customer_id,
+                )
+            except AdobeAPIError as e:
+                if e.code != STATUS_INTERNAL_SERVER_ERROR:
+                    is_valid_to_move_to_next_step = False
+                logger.warning(
```
d3rky, line 180: Please log it as exception :-)

adobe_vipm/flows/helpers.py
```diff
@@ -163,13 +169,21 @@ def __call__(self, client, context, next_step):
...
-            )
-        context.adobe_new_order_id = get_adobe_order_id(context.order)
-        logger.info(f"{context}: initialization completed.")
-        next_step(client, context)
+            try:
+                context.adobe_customer = adobe_client.get_customer(
+                    context.authorization_id,
+                    context.adobe_customer_id,
+                )
+            except AdobeAPIError as e:
```
d3rky, line 177: I think we should have following logic here

- catch exception
- log it with exception level
- finish pipeline to reschedule order for the next iteration

d3rky (changes requested):

adobe_vipm/flows/helpers.py
robcsegal: @d3rky Updated

adobe_vipm/flows/helpers.py
robcsegal: @d3rky Pushed an update.

d3rky (approved):

PR #369, opened 2025-04-08
[MPT-8676] "Value has to be a non empty array." for GET /catalog/items?in(externalIds.vendor,()) in transfer validation

jentyk:
Fix for https://softwareone.atlassian.net/browse/MPT-8676

I commented in a few places in the review to clarify the decisions behind the implementation of the fix.

I've been forced by SonarQube to do a refactoring. Since I'm very new, I played it safe and just mainly broke down the function I fixed into helper functions. Please let me know if I should continue working on this and deliver a better refactoring, but it will require understanding the business logic.

### DIFF
--- a/adobe_vipm/flows/validation/transfer.py
+++ b/adobe_vipm/flows/validation/transfer.py
@@ -77,7 +77,7 @@ def get_prices(order, commitment, adobe_skus):
         return get_prices_for_skus(product_id, currency, adobe_skus)
 
 
-def update_order_lines(
+def _update_order_lines(
     order, adobe_items, prices, items_map, quantity_field, order_error, returned_skus
 ):
     for adobe_line in adobe_items:
@@ -92,7 +92,10 @@ def update_order_lines(
                     item_sku=get_partial_sku(adobe_line["offerId"]),
                 ),
             )
-            return True, order
+            order_error = True
+
+            return order_error, order
+
         current_line = get_order_line_by_sku(
             order, get_partial_sku(adobe_line["offerId"])
         )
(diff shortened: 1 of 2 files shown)


adobe_vipm/flows/validation/transfer.py
```diff
@@ -141,30 +141,47 @@ def add_lines_to_order(
...
-
-    if is_transferred:
-        if are_all_transferring_items_expired(adobe_items_without_one_time_offers):
-            # If the order already has items and all the items on Adobe to be migrated are
-            # expired, the user can add, edit or delete the expired subscriptions
-            if len(order["lines"]):
-                return False, order
-
-            adobe_items = adobe_items_without_one_time_offers
+    if adobe_items:
```
jentyk, line 144: I assumed that `adobe_items` may be an empty list in certain cases. I need someone who knows the system better to validate that assumption.

tests/flows/validation/test_transfer.py
```diff
@@ -416,9 +417,18 @@ def test_validate_transfer_no_items(
...
+        spec=requests.Response, wraps=requests.Response
+    )
+    http_error.response.status_code = 400
+    http_error.response.json.return_value = {
+        "rql_validation": ["Value has to be a non empty array."]
+    }
+
     mocker.patch(
         "adobe_vipm.flows.validation.transfer.get_product_items_by_skus",
-        return_value=[],
```
jentyk, line 421: I found a test already covering a case of a transfer with no items, so I did not add additional tests. However, I came to conclusion that there was a wrong assumption made in preparation phase of this test. In reality the mock of `get_product_items_by_skus` library function should throw a HTTPError, in case of validation of an order with no items, not just return an empty list, as it is what happened in reality in such a scenario. After implementing the fix it will not happen anyway, but it reflects real scenarios better and will safeguard against future modifications which could introduce a regression around logic implemented here.

tests/flows/validation/test_transfer.py
```diff
@@ -435,6 +440,7 @@ def test_validate_transfer_no_items(
     assert param["error"] == ERR_ADOBE_MEMBERSHIP_ID_EMPTY.to_dict()
     assert param["constraints"]["hidden"] is False
     assert param["constraints"]["required"] is True
+    assert not get_product_items_by_skus_mock.called
```
d3rky, line 443: I would prefer here to use function `assert_not_called`

d3rky (approved):

tests/flows/validation/test_transfer.py
jentyk: Done!

PR #373, opened 2025-04-10
MPT-8440 Non active GC deployments are migrated from Adobe

alephsur:


adobe_vipm/adobe/client.py
```diff
@@ -1,3 +1,4 @@
+# Standard library imports
```
d3rky, line 1: There is no need for such kind of comments, let ruff do the staff here

adobe_vipm/adobe/client.py
```diff
@@ -1063,9 +1061,31 @@ def get_customer_deployments(
         response.raise_for_status()
         return response.json()
 
+    def get_customer_deployments_by_status(
```
d3rky, line 1064: I would say that it should be a util function that either filters out inactive deployments or splits them by active/nonactive. The thing is that if you put it to the client it should be a http query filter in that case. But since it is important to log what was removed here, I think better to have an util function that splits deployments by status

adobe_vipm/adobe/client.py
```diff
@@ -1063,9 +1061,31 @@ def get_customer_deployments(
...
+
+        for deployment in customer_deployments["items"]:
+            if deployment["status"] == status:
+                deployments_active.append(deployment)
+            else:
+                deployments_removed.append(deployment.get('deploymentId',''))
+
+        logger.info(f"Deployments removed by status: {deployments_removed}")
 
+        return {
```
d3rky, line 1083: It should be a split funtion that returns two lists

tests/adobe/test_client.py
```diff
@@ -2533,3 +2533,130 @@ def test_get_returnable_orders_by_sku_no_renewal_for_period(
...
+        "authorization_uk"
+    ]
+    customer_id = "a-customer-id"
+
+    client, authorization, api_token = adobe_client_factory()
+
+    expected_response = {
+        "totalCount": 1,
+        "items": [
+            {
```
d3rky, line 2554: please, add here non-active deployments, to be sure, that non of filter applies

tests/adobe/test_client.py
```diff
@@ -2533,3 +2533,130 @@ def test_get_returnable_orders_by_sku_no_renewal_for_period(
...
+            ),
+        ],
+    )
+
+    active_deployments = client.get_customer_deployments_by_status(
+        authorization_uk, customer_id, status
+    )
+
+    assert len(active_deployments) == 2
+    assert active_deployments.get("items",[])[0]["deploymentId"] == "deployment-1"
```
d3rky, line 2661: Please, compare full lists here, not just two fields

d3rky (changes requested):

PR #383, opened 2025-05-06
MPT-9711 Configuration orders support for Adobe Extension

alephsur:


adobe_vipm/flows/fulfillment/configuration.py
```diff
@@ -0,0 +1,134 @@
...
+                customer_id,
+                subscription_vendor_id)
+
+            try:
+                if not adobe_subscription:
+                    logger.info(
+                        f"No matching Adobe subscription found for vendor_id "
+                        f"{subscription_vendor_id} in order {context.order['id']}"
+                    )
+                    raise SubscriptionNotFoundError("No matching Adobe subscription found")
```
d3rky, line 72: Looks really strange that you raise this exception and handle it in the same block... Ideally it should be either inside the `get_subscription` or you should run the function

What I propose - is to move

1. Move this block outside the try
2. create a separate function for exception handling
3. use this function both in except block and in `if`
4. Do not raise exception

adobe_vipm/flows/fulfillment/configuration.py
```diff
@@ -0,0 +1,134 @@
...
+                )
+                update_success_subscriptions.append(
+                    {
+                        "subscription_vendor_id": subscription_vendor_id,
+                        "auto_renewal": desired_auto_renewal,
+                        "quantity": desired_quantity
+                    })
+            except AdobeError as e:
+                error_message = f"Error updating Adobe subscription {subscription_vendor_id}: {e}"
+                logger.error(error_message)
```
d3rky, line 101: Mm... keep in mind that every logged error or exception should be processed later on dashboard. Is it really something that the dashboard master should be aware of?

adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -196,7 +198,6 @@ def switch_order_to_failed(client, order, status_notes):
     order = fail_order(
         client,
         order["id"],
-        status_notes,
```
d3rky, line 199: Could you please leave bumping of sdk version on my side. The thing is that fix with fail order should be delivered in release, and your changes not :-(

adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -271,7 +272,7 @@ def handle_retries(client, order, adobe_order_id, adobe_order_type="NEW"):
     )
     reason = f"Due date is reached ({due_date_str})."
     fail_order(
-        client, order["id"], reason, ERR_VIPM_UNHANDLED_EXCEPTION.to_dict(error=reason)
+        client, order["id"], ERR_VIPM_UNHANDLED_EXCEPTION.to_dict(error=reason)
```
d3rky, line 275: same here

adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -984,6 +1012,18 @@ def __call__(self, client, context, next_step):
         next_step(client, context)
 
 
+class CompleteConfigurationOrder(CompleteOrder):
```
d3rky, line 1015: I would ask you to have a little bit for flexibility in the `CompleteOrder` step here

1. Modify the complete order step, that if template_name is None - it should call some `get_template_name` function in the `__call__`.
2. This function by default should raise the exception if template_name is None
3. Make template_name as an optional parameter in `CompleteOrder`
4. Here the only thing that you should do is to define the `get_template_name` function

Overriding the whole `__call__` is too much as for me

adobe_vipm/flows/fulfillment/base.py
```diff
@@ -41,6 +43,8 @@ def fulfill_order(client, order):
             fulfill_transfer_order(client, order)
         elif is_change_order(order):
             fulfill_change_order(client, order)
+        elif is_configuration_order(order):
```
d3rky, line 46: I think it is already a time to replace it with `match` case :-)

adobe_vipm/flows/utils.py
```diff
@@ -702,6 +706,34 @@ def notify_missing_prices(agreement_id, missing_skus, product_id, currency, comm
...
+    message = (
+        f"{error_message}\n\n"
+        f"The order **{order_id}**\n\n"
+        f"has failed changing the auto-renewal status\n\n "
+        f"- Product ID: {product_id}\n\n"
+    )
+
+    if updated_subscriptions:
+        message += "The following subscriptions has been updated and rolled back:\n"
+        for subscritpion in updated_subscriptions:
```
d3rky, line 728: typo in `subscritpion`

adobe_vipm/flows/utils.py
```diff
@@ -702,6 +706,34 @@ def notify_missing_prices(agreement_id, missing_skus, product_id, currency, comm
...
+            message += f"  - {subscritpion["subscription_vendor_id"]}\n"
+
+    send_exception(
+        "Subscription not updated",
```
d3rky, line 732: I would suggest to send something meaningful here as a title of the message. At least to add order configuration context

adobe_vipm/flows/utils.py
d3rky, line 728: you can replace it with list comprehension

```
if updated_subscriptions:
  message += "The following subscriptions has been updated and rolled back:\n"
  message += "".join(f"   - {sub['subscripton_vendor_id']}\n" for sub in updated_subscriptions])
```

tests/flows/fulfillment/test_configuration.py
```diff
@@ -0,0 +1,409 @@
...
+    step(mocked_client, context, mocked_next_step)
+
+    mocked_adobe_client.get_subscription.assert_called_once_with(
+        order["authorization"]["id"],
+        '',
+        subscriptions[0]["externalIds"]["vendor"],
+    )
+    mocked_adobe_client.update_subscription.assert_called_once_with(
+        order["authorization"]["id"],
+        '',
```
d3rky, line 52: Why it is empty?

adobe_vipm/flows/fulfillment/shared.py
d3rky, line 1015: I would ask you to have a little bit for flexibility in the `CompleteOrder` step here

1. Modify the complete order step, that if template_name is None - it should call some `get_template_name` function in the `__call__`.
2. This function by default should raise the exception if template_name is None
3. Make template_name as an optional parameter in `CompleteOrder`
4. Here the only thing that you should do is to define the `get_template_name` function

Overriding the whole `__call__` is too much as for me

adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -628,6 +644,18 @@ def __call__(self, client, context, next_step):
             send_email_notification(client, context.order)
         next_step(client, context)
 
+class ConfigurationStartOrderProcessing(StartOrderProcessing):
```
d3rky, line 647: Same as CompleteOrder

tests/flows/fulfillment/test_configuration.py
```diff
@@ -0,0 +1,409 @@
...
+    )
+    adobe_sub_2 = adobe_subscription_factory(
+        subscription_id="a-sub-id_2",
+        renewal_quantity=5,
+        autorenewal_enabled=False,
+    )
+
+    mocked_adobe_client = mocker.MagicMock()
+
+    def get_subscription_side_effect(auth_id, cust_id, vendor_id):
```
d3rky, line 321: that's not how it should be defined, you can define a list of side effects here, they are going to be picked up one by one for each call

tests/flows/fulfillment/test_configuration.py
```diff
@@ -0,0 +1,409 @@
...
+
+    context = Context(order=order, product_id='PRD-1111-1111')
+
+    step = SubscriptionUpdateAutoRenewal()
+    step(mocked_client, context, mocked_next_step)
+
+    assert mocked_adobe_client.update_subscription.call_count == 3  # 2 updates + 1 rollback
+
+    first_update_call = mocked_adobe_client.update_subscription.call_args_list[0]
+    assert first_update_call[0][2] == adobe_sub_1["subscriptionId"]
```
d3rky, line 361: please compare the whole calls instead of particular fields here

d3rky (changes requested):

adobe_vipm/flows/fulfillment/shared.py
alephsur: I have to make this changes because the tests were failing when the PR was analyzed. The new version of the SDK was used in the test. Looks like that the poetry.lock file is not used in the deployments. That's why I make this changes to pass the test correctly.

adobe_vipm/flows/fulfillment/configuration.py
```diff
@@ -0,0 +1,136 @@
...
+            adobe_client.update_subscription(
+                auth_id, customer_id, adobe_sub["subscriptionId"],
+                auto_renewal=desired, quantity=qty
+            )
+            updated.append({
+                "subscription_vendor_id": subscription_vendor_id,
+                "auto_renewal": desired,
+                "quantity": qty})
+            logger.info(f"Updated subscription {subscription_vendor_id}: "
+                        f"autoRenew={desired}, qty={qty}")
```
svazquezco, line 94: I’d suggest moving this out of the try/except, since update_subscription is the only method that can raise an AdobeError. It simplifies the code because you don’t need to wonder which line might raise the exception. This makes it easier to read and maintain in the future.

adobe_vipm/flows/fulfillment/configuration.py
```diff
@@ -0,0 +1,136 @@
...
+
+        updated = []
+        for subscription in context.order["subscriptions"]:
+            try:
+                self._process_one_subscription(adobe_client, auth_id, customer_id,
+                                               subscription, updated)
+            except SubscriptionUpdateError as e:
+                self._handle_subscription_error(client, adobe_client, auth_id, customer_id,
+                                                updated, context, str(e))
+                return
```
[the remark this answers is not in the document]
svazquezco: Ups! I deleted the message just as you were replying. I finally found the flow in the test with the rollback. The file was collapsed. Thank for the reply, and sorry for bothering you!

adobe_vipm/flows/fulfillment/base.py
```diff
@@ -33,16 +35,19 @@ def fulfill_order(client, order):
...
-        elif is_transfer_order(order):
-            order = populate_order_info(client, order)
-            order = start_processing_attempt(client, order)
-            fulfill_transfer_order(client, order)
-        elif is_change_order(order):
-            fulfill_change_order(client, order)
-        elif is_termination_order(order):  # pragma: no branch
-            fulfill_termination_order(client, order)
+        match order:
+            case _ if is_purchase_order(order):
```
d3rky, line 39: In that case I would prefer to use
```
match order["type"]:
  case MPT_ORDER_STATUS_PROCESSING
```

adobe_vipm/flows/constants.py
```diff
@@ -174,6 +175,29 @@
 ERR_ADOBE_UNEXPECTED_ERROR = "Adobe returned an unexpected error"
 
 
+MPT_ORDER_STATUS_PROCESSING = "Processing"
```
d3rky, line 178: Why did you copy it here if you have same on lines 304?

adobe_vipm/flows/fulfillment/configuration.py
```diff
@@ -0,0 +1,141 @@
...
+    pipeline.run(client, context)
+
+
+
+class SubscriptionUpdateAutoRenewal(Step):
+    def __call__(self, client, context, next_step):
+        """
+        Updates the auto renewal status of a subscription.
+        """
+        logger.info(f'Start processing {context.order["type"]} order {context.order["id"]}')
```
d3rky, line 52: Same log as for `fulfill_configuration_order`. Why do you need it?

adobe_vipm/flows/fulfillment/configuration.py
```diff
@@ -0,0 +1,141 @@
...
+                                               subscription, updated)
+            except SubscriptionUpdateError as e:
+                self._handle_subscription_error(client, adobe_client, auth_id, customer_id,
+                                                updated, context, str(e))
+                return
+
+        next_step(client, context)
+
+
+    def _process_one_subscription(self, adobe_client, auth_id, customer_id, subscription, updated):
```
d3rky, line 70: `_process_subscription` would be better here :-)

and I would prefer to pass here context. You can extract all the parameters from context

d3rky (changes requested):

adobe_vipm/flows/fulfillment/base.py
alephsur: done

adobe_vipm/flows/fulfillment/configuration.py
alephsur: log removed

adobe_vipm/flows/fulfillment/configuration.py
alephsur: function name and parameters changed.

adobe_vipm/flows/fulfillment/base.py
```diff
@@ -33,16 +32,20 @@ def fulfill_order(client, order):
...
-            order = populate_order_info(client, order)
-            order = start_processing_attempt(client, order)
-            fulfill_transfer_order(client, order)
-        elif is_change_order(order):
-            fulfill_change_order(client, order)
-        elif is_termination_order(order):  # pragma: no branch
-            fulfill_termination_order(client, order)
+        match order["type"]:
+            case constants.ORDER_TYPE_PURCHASE:
+                if not is_transfer_order(order):
```
d3rky, line 37: It would be better not to use `not` in `if` statements. Just easier to read

d3rky (approved):

svazquezco (approved):

PR #386, opened 2025-05-08
MPT-7834 Migrating three year committed produced empty items

alephsur:
MPT-7834 refactor the validate function

### DIFF
--- a/tests/flows/validation/test_transfer.py
+++ b/tests/flows/validation/test_transfer.py
@@ -19,6 +19,7 @@ from adobe_vipm.flows.constants import (
     ERR_ADOBE_MEMBERSHIP_ID_INACTIVE_ACCOUNT,
     ERR_ADOBE_MEMBERSHIP_ID_ITEM,
     ERR_ADOBE_MEMBERSHIP_NOT_FOUND,
+    ERR_ADOBE_MEMBERSHIP_PROCESSING,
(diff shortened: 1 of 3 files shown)


tests/flows/validation/test_transfer.py
```diff
@@ -1704,3 +1705,164 @@ def test_validate_transfer_with_one_line_items(
...
+    )
+
+    # Assert
+    assert has_errors is True
+    param = get_ordering_parameter(validated_order, PARAM_MEMBERSHIP_ID)
+
+    assert param["error"] == ERR_ADOBE_MEMBERSHIP_PROCESSING.to_dict(
+        membership_id=mocked_transfer.membership_id,
+        error='5112 - Test error',
+    )
```
jentyk, line 1868: These 3 tests are quite similar can we use `@pytest.mark.parametrize` here?

jentyk (approved):
LGTM - based on my VERY limited business logic understanding.
Strongly advised one more approval

tests/flows/validation/test_transfer.py
alephsur: Yes, You are right. I will implement the changes to use only one test with @pytest.mark.parametrize. Thank you for the advice

d3rky (approved):

PR #391, opened 2025-05-16
MPT-9841 Adobe pricing in Global Customer location agreements do not …

alephsur:
…honour discount level of main agreement


adobe_vipm/flows/global_customer.py
```diff
@@ -634,6 +659,51 @@ def process_agreement_deployment(
         agreement_deployment.save()
 
 
+def get_sku_price(adobe_customer, adobe_subscription, product_id, deployment_currency):
+    """
+    Obtiene el precio del SKU considerando el nivel de descuento y el compromiso 3YC.
```
d3rky, line 664: Would be nice to have comments in english ;-)

adobe_vipm/flows/global_customer.py
```diff
@@ -634,6 +659,51 @@ def process_agreement_deployment(
         agreement_deployment.save()
 
 
+def get_sku_price(adobe_customer, adobe_subscription, product_id, deployment_currency):
```
d3rky, line 662: I didn't get this function. At least same logic should already exists, because when you sync agreements and subscrpitions the price there is correct and honour 3yc

adobe_vipm/flows/global_customer.py
```diff
@@ -427,6 +441,7 @@ def create_gc_agreement_subscription(
     gc_agreement_id,
     buyer_id,
     item,
+    prices=None
```
d3rky, line 444: Didn't get this `None`. are there any cases when we shouldn't create gc subscription without proper pricing?

d3rky (changes requested):

adobe_vipm/flows/global_customer.py
alephsur: I refactor the sync agreements to use the same method in both places.

adobe_vipm/flows/global_customer.py
```diff
@@ -467,7 +475,9 @@ def create_gc_agreement_subscription(
             ]
         },
         "externalIds": {"vendor": adobe_subscription["subscriptionId"]},
-        "lines": [{"quantity": adobe_subscription["currentQuantity"], "item": item}],
+        "lines": [{"quantity": adobe_subscription["currentQuantity"], "item": item,
```
d3rky, line 478: What type is `prices` here?. It looks like it is just a price of the item and that it.
In that case why do you need so complex logic on merging dicts???
Why not to:

```
"lines": [
  {
    "quantity": adobe_subscription["currentQuantity"],
    "item": item,
    "price": {"unitPP": price}
  }]
``` 
???

d3rky (changes requested):

d3rky (approved):

PR #398, opened 2025-05-21
MPT-10181 Broken link(s) to an image in notification email from Adobe extension

jentyk:
When there is no icon property available for the Client email shows no icon but:
image

We want something more easthetically pleasing.


jentyk:
This PR, in case of missing icon property for the Client, makes email header looking like following:
image

d3rky (approved):

PR #399, opened 2025-05-23
MPT-10404 Termination Orders and Returns

alephsur:
This PR has a direct dependency with the a bug reported to the plaftorm team to terminate the entire agreement. The validation process is not working for an agreement but it does for the subscriptions.
https://softwareone.atlassian.net/browse/MPT-10436?linkSource=email


adobe_vipm/flows/validation/termination.py
```diff
@@ -0,0 +1,48 @@
+import logging
+
+from adobe_vipm.adobe.client import get_adobe_client
+from adobe_vipm.flows.constants import ERR_INVALID_TERMINATION_ORDER_QUANTITY
+from adobe_vipm.flows.context import Context
+from adobe_vipm.flows.helpers import SetupContext, ValidateDownsizes3YC
+from adobe_vipm.flows.pipeline import Pipeline, Step
+from adobe_vipm.flows.utils import (
+    _validate_subscription_and_returnable_orders,
```
jentyk, line 9: The same as fulfillment, if it is imported, it should not be private.

tests/flows/fulfillment/test_termination.py
```diff
@@ -384,3 +414,134 @@ def test_switch_autorenewal_off_error_updating_autorenew(
...
+    order = order_factory(
+        lines=lines_factory(
+            quantity=0,
+            old_quantity=7,
+        )
+    )
+    adobe_customer = adobe_customer_factory()
+    sku = order["lines"][0]["item"]["externalIds"]["vendor"]
+
+    mocked_adobe_client = mocker.MagicMock()
```
jentyk, line 437: The same as above, I think it would be good to spec it.

tests/flows/fulfillment/test_termination.py
```diff
@@ -384,3 +414,134 @@ def test_switch_autorenewal_off_error_updating_autorenew(
...
+
+    mocked_switch_to_failed = mocker.patch(
+        "adobe_vipm.flows.fulfillment.termination.switch_order_to_failed",
+    )
+
+    mocker.patch(
+        "adobe_vipm.flows.fulfillment.termination.get_adobe_client",
+        return_value=mocked_adobe_client,
+    )
+    mocked_client = mocker.MagicMock()
```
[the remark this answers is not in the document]
alephsur: This must be refactored in all the test that we have in the extension. I think that we need to create a task only to correct this in all the tests in the extension.

adobe_vipm/flows/validation/base.py
```diff
@@ -39,18 +39,23 @@ def validate_order(client, order):
...
-            order = populate_order_info(client, order)
-            order = reset_ordering_parameters_error(order)
-            order = reset_order_error(order)
-            has_errors, order = validate_transfer(client, adobe_client, order)
+        match order["type"]:
+            case constants.ORDER_TYPE_PURCHASE:
+                if is_transfer_order(order) and is_transfer_validation_enabled(
+                    order
+                ):
+                    adobe_client = get_adobe_client()
```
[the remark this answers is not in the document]
jentyk: True, but my comment was more about reuse of HTTP session. Anyway, you can add TODO if you want, and I resolve this thread anyway.

adobe_vipm/flows/utils.py
```diff
@@ -1076,3 +1078,64 @@ def get_address(data):
...
+
+def validate_subscription_and_returnable_orders(
+    adobe_client,
+    context,
+    line,
+    sku,
+    return_orders=None
+):
+    """
+    Validates if the subscription is active and has valid returnable orders.
```
d3rky, line 1105: Please, add here definition of what is `valid returnable orders`

adobe_vipm/flows/utils.py
```diff
@@ -1076,3 +1078,64 @@ def get_address(data):
...
+        sku,
+        context.adobe_customer["cotermDate"],
+        return_orders=return_orders,
+    )
+
+    if not returnable_orders:
+        return False, []
+
+    if not has_valid_returnable_quantity(line, returnable_orders):
+        return False, None
```
d3rky, line 1139: What is the difference between returning

```
False, None
vs
False, []
```

adobe_vipm/flows/utils.py
```diff
@@ -1076,3 +1078,64 @@ def get_address(data):
...
+        context: The context object
+        line: The order line to validate
+        sku: The SKU to validate
+        return_orders: Optional return orders to pass to get_returnable_orders_by_sku
+
+    Returns:
+        tuple: (is_valid, returnable_orders)
+        - is_valid: True if validation passed, False otherwise
+        - returnable_orders: List of returnable orders if validation passed, None otherwise
+    """
```
d3rky, line 1118: I think we should add a little bit of logging here, espacially for all the returning branches, otherwise to understand what happend in production without it would be painfull :-(

adobe_vipm/flows/utils.py
alephsur: done

adobe_vipm/flows/utils.py
d3rky, line 1: I would like to ask you to spend some time on making this file smaller by decoupling it to multiple. Please, otherwise it just becomes bigger and bigger

d3rky (changes requested):

adobe_vipm/flows/utils.py
alephsur: I will make the refactor for this file.

alephsur: done

d3rky (approved):

PR #400, opened 2025-05-29
MPT-10651 3YC Price not honored on change

alephsur:


adobe_vipm/flows/helpers.py
```diff
@@ -345,3 +349,90 @@ def is_return_order_created(context):
...
+            next_step(client, context)
+            return
+
+        actual_skus = self._get_actual_skus(context)
+        prices = self._get_prices_for_skus(context, actual_skus)
+
+        print("RRRRRRRRRR12")
+        print("RRRRRRRRRR12")
+        print(actual_skus)
+        print("RRRRRRRRRR12")
```
d3rky, line 366: :-(

d3rky (changes requested):

adobe_vipm/flows/helpers.py
alephsur: print("sorry"), It is removed.

d3rky: > Could we add https://docs.astral.sh/ruff/rules/print/?

In my todo list

ruben-sebrango (approved):

PR #404, opened 2025-06-05
MPT-10517 Delayed Order Improvements

alephsur:


adobe_vipm/flows/utils/date.py
```diff
@@ -93,10 +94,19 @@ def is_renewal_window_open(order):
...
         datetime.fromisoformat(coterm_date) - timedelta(days=LAST_TWO_WEEKS_DAYS)
     ).date()
 
     return date.today() >= last_two_weeks
+
+def is_coterm_date_in_last_24_hours(order):
+    if not get_coterm_date(order):
+        return False
+    coterm_date = datetime.fromisoformat(get_coterm_date(order)).date()
+    pacific_tz = pytz.timezone('America/Los_Angeles')
```
d3rky, line 108: Could you also drop a comment here, why we use LA timezone here. For the future :-)

d3rky (changes requested):

adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -605,16 +607,11 @@ def __init__(self, template_name):
         self.template_name = template_name
 
     def __call__(self, client, context, next_step):
-        template_name = (
-            self.template_name
-            if not is_renewal_window_open(context.order)
```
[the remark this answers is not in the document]
alephsur: No, that function was only used to assign the correct template for the delayed order.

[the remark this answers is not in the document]
d3rky: Can we remove this function in that case?

tests/flows/fulfillment/test_configuration.py
```diff
@@ -414,13 +414,15 @@ def test_fulfill_configuration_order(mocker):
 
     fulfill_configuration_order(mocked_client, mocked_order)
 
-    assert len(mocked_pipeline_ctor.mock_calls[0].args) == 5
+    assert len(mocked_pipeline_ctor.mock_calls[0].args) == 7
 
     assert isinstance(mocked_pipeline_ctor.mock_calls[0].args[0], Step)
```
d3rky, line 419: Could you fix here also

1. Compare with the list
2. Put concrete steps that you are using, not just basic class `Step`

adobe_vipm/flows/validation/termination.py
```diff
@@ -39,6 +43,8 @@ def validate_termination_order(client, order):
     pipeline = Pipeline(
         SetupContext(),
         ValidateDuplicateLines(),
+        SetOrUpdateCotermNextSyncDates(),
+        ValidateRenewalWindow(True),
```
d3rky, line 47: Please for boolean arguments use keyword notation

```
ValidateRenewalWindow(is_validation=True)
```

just for readability

adobe_vipm/flows/validation/termination.py
```diff
@@ -39,6 +43,8 @@ def validate_termination_order(client, order):
     pipeline = Pipeline(
         SetupContext(),
         ValidateDuplicateLines(),
+        SetOrUpdateCotermNextSyncDates(),
+        ValidateRenewalWindow(is_validation=True),
```
d3rky, line 47: Could you please write down and fix it somewhere on next PRs, I would prefer to make this step with an hour parameter for time limit

d3rky (approved):

PR #406, opened 2025-06-09
MPT-10817 Synchronization triggered on order completion

jentyk:


adobe_vipm/flows/fulfillment/base.py
```diff
@@ -46,6 +49,7 @@ def fulfill_order(client, order):
                 fulfill_configuration_order(client, order)
             case constants.ORDER_TYPE_TERMINATION:
                 fulfill_termination_order(client, order)
+        sync_agreement(client, get_agreement(client, order["agreement"]["id"]), dry_run=False)
```
d3rky, line 52: Why do you prefer to have logic here, instead of having an explicit step and add this step all the pipelines?

jentyk: I'm do not have a very strong opinion on this one, but see two reasons why it could be better to do it this way.

1. Arguably, syncing does not seem to me a part of any order processing flow. Of course, we need to update information in our system afterwards, but it is not one of the steps we carry on through the customer API.
2. The requirement is to sync after every order processing procedure regarding the result (errors etc.). Placing the sync call at this level clearly illustrates meeting requirement in code logic. It is clear from the code it will be executed regardless what happened with the order.

Actually you just helped me to find a bug in my code 😱 . it should be placed in `finally` clause. I am going to change it now.

jentyk: In case of Agreement has processing subscriptions sync is not taking place - sync.py:321 . I've seen this when I was doing manual integrations tests. But let me know if you still want to make it a step

d3rky (approved):

albertsola (approved):

PR #416, opened 2025-06-18
MPT-11063 Add lastSyncDate parameter to fulfillment constraints

jentyk:


adobe_vipm/flows/sync.py
```diff
@@ -66,178 +67,158 @@ def sync_agreement_prices(mpt_client, agreement, dry_run, adobe_client, customer
...
 
-            line_id = subscription["lines"][0]["id"]
-            lines = [
-                {
-                    "price": {"unitPP": prices[actual_sku]},
-                    "id": line_id,
-                }
-            ]
+        if adobe_subscription["status"] == STATUS_SUBSCRIPTION_TERMINATED:
+            logger.info(f"Skipping subscription {subscription['id']}. It is terminated")
```
d3rky, line 103: I know that's not your code, but worth to add it. Could you please add to the log message, that it is terminated on Adobe :-)

adobe_vipm/flows/sync.py
```diff
@@ -355,8 +327,22 @@ def sync_agreement(mpt_client, agreement, dry_run):
...
         logger.exception(f"Error synchronizing agreement {agreement["id"]}: {e}")
-        notify_agreement_unhandled_exception_in_teams(
-            agreement["id"], traceback.format_exc()
+        notify_agreement_unhandled_exception_in_teams(agreement["id"], traceback.format_exc())
+    else:
+        _update_last_sync_date(mpt_client, agreement, dry_run)
+
+
+def _update_last_sync_date(mpt_client: MPTClient, agreement: dict, dry_run: bool) -> None:
+    if not dry_run:
```
d3rky, line 336: I would recommend putting it out of the function :-) In the `else` branch. Just to have everything in one place, why this parameter can't be updated. So we have 2 reasons:
1. Dry run
2. Failure

Otherwise these 2 conditions a hidden in 2 different places

robcsegal (approved):

PR #431, opened 2025-07-02
MPT-11615 Make the improvements for the quality gate

alephsur:


adobe_vipm/adobe/constants.py
```diff
@@ -77,15 +79,18 @@
...
-STATUS_3YC_DECLINED = "DECLINED"
-STATUS_3YC_COMMITTED = "COMMITTED"
-STATUS_3YC_ACTIVE = "ACTIVE"
-STATUS_3YC_REQUESTED = "REQUESTED"
-STATUS_3YC_NONCOMPLIANT = "NONCOMPLIANT"
-STATUS_3YC_EXPIRED = "EXPIRED"
-OFFER_TYPE_LICENSE = "LICENSE"
-OFFER_TYPE_CONSUMABLES = "CONSUMABLES"
+
+class ThreeYearCommitmentStatus(Enum):
```
d3rky, line 83: I think you can use `StrEnum` here and avoid using `.value` everywhere

d3rky (changes requested):

adobe_vipm/adobe/constants.py
jentyk: This is great! thanks for that. I have one suggestion to use @unique. Also could you please add TEMPORARY group replacing line adobe_vipm/flows/sync.py:56. What I mean is something like that:

class ThreeYearCommitmentStatus(Flag):
    ACCEPTED = auto()
    DECLINED = auto()
    COMMITTED = auto()
    ACTIVE = auto()
    REQUESTED = auto()
    NONCOMPLIANT = auto()
    EXPIRED = auto()
    
    TEMPORARY = REQUESTED | ACCEPTED
    
    def __getattribute__(self, name):
        if name == 'value':
            return self.name
        
        return super().__getattribute__(name)

d3rky (approved):

svazquezco (approved):

adobe_vipm/flows/validation/transfer.py
```diff
@@ -70,7 +70,10 @@ def get_prices(order, commitment, adobe_skus):
     product_id = order["agreement"]["product"]["id"]
     if (
         commitment
-        and commitment["status"] in (STATUS_3YC_COMMITTED, "ACTIVE")
+        and commitment["status"] in (
+            ThreeYearCommitmentStatus.COMMITTED,
```
jentyk, line 74: Here also group coul dbe created on the enum. Would be more concise and readable

adobe_vipm/flows/sync.py
```diff
@@ -53,7 +50,10 @@
     notify_missing_prices,
 )
 
-TEMP_3YC_STATUSES = (STATUS_3YC_REQUESTED, STATUS_3YC_ACCEPTED)
+TEMP_3YC_STATUSES = (
+    ThreeYearCommitmentStatus.REQUESTED,
```
jentyk, line 54: Also here...

adobe_vipm/flows/sync.py
```diff
@@ -87,7 +87,10 @@ def sync_agreement_prices(mpt_client, agreement, dry_run, adobe_client, customer
     commitment_start_date = None
     if (
         commitment
-        and commitment["status"] in (STATUS_3YC_COMMITTED, STATUS_3YC_ACTIVE)
+        and commitment["status"] in (
+            ThreeYearCommitmentStatus.COMMITTED,
```
jentyk, line 91: And here that group could be used

adobe_vipm/adobe/constants.py
d3rky: Lukasz, I think that's overcomplicating the solution, since an enum represents a set of possible values. So, if you want to do shortcuts, just do it :-)

```
TEMPORARY_3YC_STATUSES = (ThreeYearCommitmentStatus.ACCEPTED, ThreeYearCommitmentStatus. REQUESTED)
```

PR #436, opened 2025-07-03
Mpt 11615 make the improvements for the quality gate

alephsur:


adobe_vipm/flows/benefits.py
```diff
@@ -120,14 +120,16 @@ def check_3yc_commitment_request(mpt_client, is_recommitment=False):
...
+                ThreeYearCommitmentStatus.EXPIRED,
+                ThreeYearCommitmentStatus.NONCOMPLIANT,
             ):
                 agreement_link = urljoin(
                     settings.MPT_PORTAL_BASE_URL,
                     f"/commerce/agreements/{agreement['id']}",
                 )
                 send_warning(
-                    f"3YC {request_type_title.capitalize()} Request {str(status).capitalize()}",
+                    f"3YC {request_type_title.capitalize()} Request {status.capitalize()}",
```
albertsola, line 132: I would remove the capitalize() from status.

adobe_vipm/adobe/mixins/order.py
```diff
@@ -93,14 +93,14 @@ def build_line_item(item):
                 line_item["currencyCode"] = item["currencyCode"]
             return line_item
 
-        line_items = [
+        lineItems = [
```
d3rky, line 96: I don't that that you should revert this line

tests/conftest.py
```diff
@@ -1779,7 +1779,7 @@ def mock_gunicorn_logging_config():
 
 @pytest.fixture()
 def mock_wrap_event():
-    return Event("evt-id", "orders", {"id": "ORD-1111-1111"})
+    return Event("evt-id", "orders", {"id": "ORD-1111-1111-1111"})
```
jentyk, line 1782: Just to be sure, is this an intentional change?

adobe_vipm/flows/benefits.py
jentyk: Good idea, but maybe in a separate commit or next time.

alephsur: removed

PR #437, opened 2025-07-03
MPT-11668 3YC parameters are not correctly update in complete order

alephsur:


ruben-sebrango (approved):

adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -611,7 +611,11 @@ def coterm_and_next_sync_update_if_needed(self, context, coterm_date, next_sync)
     def commitment_update_if_needed(self, context):
         if not context.adobe_customer:
             return False
-        commitment = get_3yc_commitment_request(context.adobe_customer)
+        commitment = (
+            get_3yc_commitment_request(context.adobe_customer)
```
jentyk, line 615: Just to confirm, `get_3yc_commitment_request` has `is_recommitment` param. Is it OK to use here only the case with its default value which is `False`?

PR #438, opened 2025-07-03
MPT-11668 3YC parameters are not correctly update in complete order

alephsur:


adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -594,7 +594,10 @@ def __call__(self, client, context, next_step):
                 needs_update = True
 
             if context.adobe_customer:
-                commitment = get_3yc_commitment_request(context.adobe_customer)
+                commitment = (
+                    get_3yc_commitment_request(context.adobe_customer)
```
jentyk, line 598: Just to confirm, get_3yc_commitment_request has is_recommitment param. Is it OK to use here only the case with its default value which is False?

d3rky (approved):

ruben-sebrango (approved):

adobe_vipm/flows/fulfillment/shared.py
alephsur: the commitment request is null when the customer gets to the committed status. This is a change in the new sandbox and production. That's why we need to consider both. Because the status ACCEPTED has to be considere in the commitment-request and COMMITED in the commitmen. Both are valids in diferent times in the process.

PR #439, opened 2025-07-03
MPT-11684 Solve sonarqube quality gate

alephsur:


adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -584,52 +584,64 @@ def __call__(self, client, context, next_step):
...
-                        "3yc_start_date": commitment["startDate"],
-                        "3yc_end_date": commitment["endDate"]
-                    })
-                params_str = ', '.join(f'{k}={v}' for k, v in updated_params.items())
-                logger.info(f"{context}: Updated parameters: {params_str}")
+                self.update_order_parameters(client, context, coterm_date, next_sync)
 
         next_step(client, context)
 
+    def coterm_and_next_sync_update_if_needed(self, context, coterm_date, next_sync):
```
jentyk, line 599: Just a suggestion, IMHO it would be clearer to start a function/method name using a verb. In this case `update_coterm_and_next_sync`, I would personally skip `if_needed` this is a detail which will be visible in the function body

d3rky (approved):

PR #440, opened 2025-07-03
MPT-11078 Update Agreement params being synced

jentyk:
This is just a refactoring. I had to break down PR into pieces as it is big.


albertsola (commented):
SonarCube find a few new issues worth addressing before doing a more in depth code review.

adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -485,10 +478,7 @@ def send_mpt_notification(mpt_client: MPTClient, order: dict) -> None:
         "api_base_url": settings.MPT_API_BASE_URL,
         "portal_base_url": settings.MPT_PORTAL_BASE_URL,
     }
-    subject = (
-        f"Order status update {order['id']} "
-        f"for {order['agreement']['buyer']['name']}"
-    )
+    subject = f"Order status update {order['id']} " f"for {order['agreement']['buyer']['name']}"
```
albertsola, line 481: I would personally prefer one single string in this cases or a + to indicate the strings are meant to be join

albertsola:
Many changes looks cosmetic and when I run ruff on your branch it looks like it reverted some of those changes:

```
 swo-adobe-vipm-extension % ruff format
65 files reformatted, 59 files left unchanged
```

@jentyk Can you double check ruff in your machine? If it is working, please let me know as I may have mine missconfigured.

adobe_vipm/flows/fulfillment/shared.py
jentyk: Ah! It's Ruff who does it! Thanks for spotting!

jentyk:
> Many changes looks cosmetic and when I run ruff on your branch it looks like it reverted some of those changes:
> 
> ```
>  swo-adobe-vipm-extension % ruff format
> 65 files reformatted, 59 files left unchanged
> ```
> 
> @jentyk Can you double check ruff in your machine? If it is working, please let me know as I may have mine missconfigured.

@albertsola when I run this on my there are of course some changes generated. Basically at some point I added to Adobe repo the same config as you guys have in AWS, just to use one standard across the team. But I did not reformat everything at once just doing it gradually when I touch a file. Do you believe it would be better to do everything at once, now? Maybe it's good idea actually. I add one more commit with the result of the formatting.

albertsola:
I would personally do one reformat of all files in one commit. Then all the changes. All subsequent pull request would be smaller and simpler.

adobe_vipm/adobe/utils.py
```diff
@@ -45,28 +45,6 @@ def join_phone_number(phone: dict) -> str:
     return f"{phone['prefix']}{phone['number']}" if phone else ""
 
 
-def get_3yc_commitment(customer_or_transfer_preview):
```
d3rky, line 48: Didn't get why do you prefer to remove it from adobe/utils.py and leave it in utils.py

Because `get_3yc_commitment_request` function is in adobe/utils.py also

Could you check in your next PRs?

PR #441, opened 2025-07-04
MPT-11489 When an order is successful at Adobe (status 1000), but the…

alephsur:
… renewal quantity cannot be updated, the MPT order fails.


jentyk (approved):
LGTM

PR #442, opened 2025-07-04
MPT-11608 Minumum licenses and consumables error message does not mat…

alephsur:
…ch requirements

### DIFF
--- a/adobe_vipm/flows/constants.py
+++ b/adobe_vipm/flows/constants.py
@@ -140,12 +140,14 @@ ERR_VIPM_UNHANDLED_EXCEPTION = ValidationError(
 
 ERR_3YC_QUANTITY_LICENSES = ValidationError(
     "VIPM0013",
-    "The `{title}` must be an integer " f"number equal to {MINQTY_LICENSES} or greater.",
+    f"The minimum number of licenses must be numeric "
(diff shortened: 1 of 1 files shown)


adobe_vipm/flows/constants.py
```diff
@@ -140,12 +140,14 @@
 
 ERR_3YC_QUANTITY_LICENSES = ValidationError(
     "VIPM0013",
-    "The `{title}` must be an integer " f"number equal to {MINQTY_LICENSES} or greater.",
+    f"The minimum number of licenses must be numeric "
+    f"and greater than or equal to ten ({MINQTY_LICENSES}).",
```
d3rky, line 144: How does it happen that tests are not affected?

d3rky (changes requested):

adobe_vipm/flows/constants.py
alephsur: Because the test is using the next:
@pytest.mark.parametrize(
    ("param_name", "factory_field", "error"),
    [
        (PARAM_3YC_LICENSES, "p3yc_licenses", ERR_3YC_QUANTITY_LICENSES),
        (PARAM_3YC_CONSUMABLES, "p3yc_consumables", ERR_3YC_QUANTITY_CONSUMABLES),
    ],
)

PR #453, opened 2025-07-09
MPT-11691 3YC Order and fullfillment parameters are not updated corre…

alephsur:
…ctly before order completion

### DIFF
--- a/adobe_vipm/flows/fulfillment/shared.py
+++ b/adobe_vipm/flows/fulfillment/shared.py
@@ -79,6 +79,7 @@ from adobe_vipm.flows.utils import (
     split_phone_number,
 )
 from adobe_vipm.flows.utils.customer import has_coterm_date
+from adobe_vipm.flows.utils.three_yc import set_adobe_3yc
 from adobe_vipm.notifications import mpt_notify
 from adobe_vipm.utils import get_3yc_commitment, get_partial_sku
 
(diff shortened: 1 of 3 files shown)


ruben-sebrango (approved):

alephsur:
> Hey, David! I was thinking a little bit about this issue/PR and I think we should leave current behavior as is. Cause 3YC and status are moved to agreement as parameters and it is a good idea to have a parameter to defer is this agreement has 3yc or not. And checkbox and status :-) Let's change the acceptance criterias here

In the current implementation for the command that process the agreements with the 3YC requested the behaviour is the new one as we can see in the code in benefits.py updating the parameters with the commitment information.

`
            request_type_param_ext_id = (
                Param.THREE_YC if not is_recommitment else Param.THREE_YC_RECOMMITMENT
            )

-----------------------

                parameters[request_type_param_phase].append(
                    {"externalId": request_type_param_ext_id, "value": None},
                )
                parameters[Param.PHASE_FULFILLMENT].extend(
                    [
                        {
                            "externalId": Param.THREE_YC_ENROLL_STATUS,
                            "value": commitment_info["status"],
                        },
                        {
                            "externalId": Param.THREE_YC_START_DATE,
                            "value": commitment_info["startDate"],
                        },
                        {
                            "externalId": Param.THREE_YC_END_DATE,
                            "value": commitment_info["endDate"],
                        },
                    ],`

d3rky (approved):

PR #455, opened 2025-07-10
MPT-11810 Solve the problem with the quality gate in sonarqube

alephsur:


ruben-sebrango (approved):

adobe_vipm/flows/migration.py
```diff
@@ -159,110 +159,138 @@ def populate_offers_for_transfer(product_id, transfer, transfer_preview):
...
+            "Adobe error received during transfer preview.",
+            "An unexpected error has been received from Adobe asking for preview "
+            f"of transfer for Membership **{transfer.membership_id}**.",
+            facts=FactsSection(
+                "Last error from Adobe",
+                {transfer.adobe_error_code: transfer.adobe_error_description},
+            ),
+            button=get_transfer_link_button(transfer),
+        )
+    return False
```
albertsola, line 190: Can we use the return early pattern?


```
if any( ..):
  ...
   return False

foo()
return False
```

albertsola (approved):

adobe_vipm/flows/migration.py
```diff
@@ -159,110 +159,138 @@ def populate_offers_for_transfer(product_id, transfer, transfer_preview):
     create_offers(product_id, offers)
 
 
+def handle_preview_error(transfer, api_err):
```
d3rky, line 162: What this function is doing - updating the transfer object properly according to the api_err. And I didn't get why do you need to return true/false here

adobe_vipm/flows/migration.py
```diff
@@ -159,110 +159,138 @@ def populate_offers_for_transfer(product_id, transfer, transfer_preview):
...
+        "Marketplace Platform configuration error during transfer.",
+        str(e),
+        facts=FactsSection(
+            "Transfer error",
+            {"AuthorizationNotFoundError": transfer.migration_error_description},
+        ),
+        button=get_transfer_link_button(transfer),
+    )
+
+def handle_transfer_creation_error(transfer, api_err):
```
d3rky, line 208: Why does some of the `hadle_*` preview functions do `save()` for transfer, but others not?

adobe_vipm/flows/migration.py
```diff
@@ -159,110 +159,138 @@ def populate_offers_for_transfer(product_id, transfer, transfer_preview):
...
+        f"transfer for Membership **{transfer.membership_id}**.",
+        facts=FactsSection(
+            "Last error from Adobe",
+            {transfer.adobe_error_code: transfer.adobe_error_description},
+        ),
+        button=get_transfer_link_button(transfer),
+    )
+
+def handle_transfer_reseller_error(transfer, e):
+    """Handle reseller errors during transfer creation"""
```
d3rky, line 228: I see some repetitive pattern here when you fail transfer, like when you fail with Adobe Error or not

adobe_vipm/flows/migration.py
```diff
@@ -159,110 +159,138 @@ def populate_offers_for_transfer(product_id, transfer, transfer_preview):
...
+        adobe_transfer = client.create_transfer(
+            transfer.authorization_uk,
+            transfer.seller_uk,
+            transfer.record_id,
+            transfer.membership_id,
+        )
+        return adobe_transfer, None
+    except AdobeAPIError as api_err:
+        handle_transfer_creation_error(transfer, api_err)
+        return None, "error"
```
d3rky, line 271: Didn't get this tuple to be returned. Why it is a string "error"?

d3rky (changes requested):

d3rky (approved):

PR #458, opened 2025-07-14
[MPT-11750] Decoupled runtime from extension and utilizing runtime from sdk

robcsegal:
Decoupled runtime from extension and utilizing runtime from sdk

https://softwareone.atlassian.net/browse/MPT-11750


ruben-sebrango (approved):

docker-compose.yml
```diff
@@ -10,10 +10,6 @@ services:
       - .:/extension
     env_file:
       - .env
-    ports:
-      - "8080:8080"
```
d3rky, line 14: Why do you need these changes?

robcsegal: Was committed after I did a rebase. Should have only been local. Docker-compose is now unchanged like it should be.

d3rky (approved):

PR #464, opened 2025-07-17
MPT-11087 add `end_of_sale` field to sku mapping and test data

jentyk:


adobe_vipm/flows/sync.py
```diff
@@ -274,8 +272,29 @@ def _get_subscriptions_for_update(
...
+            logger.info(f"Processing terminated Adobe subscription {adobe_subscription_id}.")
+            if is_sku_end_of_sale(adobe_subscription_id, today_date):
+                logger.info(
+                    "> The subscription is End Of Sale, terminating subscription."
+                    f" {subscription['id']}."
+                )
+                terminate_subscription(
+                    mpt_client,
+                    subscription["id"],
+                    "Adobe subscription status 1004.",
```
robcsegal, line 285: This string is repeated and there's a constant for this status.

jentyk: Can you point me to repetition? Also, where is this string repeated? Ic ould not find it

adobe_vipm/airtable/models.py
```diff
@@ -734,3 +738,13 @@ def get_adobe_product_by_marketplace_sku(
...
+
+
+@cache
+def get_adobe_sku(vendor_item_id):
+    return get_adobe_product_by_marketplace_sku(vendor_item_id).sku
+
+
+@lru_cache(maxsize=1900)
+def is_sku_end_of_sale(vendor_item_id: str, _age_hash_date_do_not_remove: int) -> bool:
+    return bool(get_adobe_product_by_marketplace_sku(vendor_item_id).end_of_sale)
```
d3rky, line 750: why not?

```
return get_adobe_product_by_marketplace_sku(vendor_item_id).is_end_of_sale()
```

adobe_vipm/airtable/models.py
```diff
@@ -734,3 +738,13 @@ def get_adobe_product_by_marketplace_sku(
...
     AdobeItemModel = get_sku_adobe_mapping_model(AirTableBaseInfo.for_sku_mapping())
     return AdobeItemModel.from_short_id(vendor_external_id)
+
+
+@cache
+def get_adobe_sku(vendor_item_id):
+    return get_adobe_product_by_marketplace_sku(vendor_item_id).sku
+
+
+@lru_cache(maxsize=1900)
```
d3rky, line 748: mm... not sure that lru_cache here is the best option. What if data in the table was changed?

jentyk: This method is here only to check if the item is the end of sale. I presume the value of this field will not be changing too frequently and the cache is designed to be valid only for one day. The issue would occur only in the case when the item has been synced on a specific day, then marked as "end of sale" the same day and synced, again on the same day. Let's remember the sync in many cases will be taking place only once per day. The changed will be picked up from the next day onwards, and if we do not cache this value, we will be sending requests to the Airtable over and over again with every sync the item occurs in the currently synced subscriptions.

d3rky (approved):

PR #470, opened 2025-07-23
MPT-11279 vadalidate the parameters introduced by the customer

alephsur:


albertsola (approved):

adobe_vipm/flows/utils/parameter.py
```diff
@@ -78,17 +79,40 @@ def reset_ordering_parameters_error(order):
...
+            PARAM_NEW_CUSTOMER_PARAMETERS + TRANSFER_CUSTOMER_PARAMETERS,
+            hidden=True
+        )
+        set_params_visibility([Param.MEMBERSHIP_ID], hidden=False)
+    elif agreement_value == "Transfer":
+        set_params_visibility(PARAM_NEW_CUSTOMER_PARAMETERS, hidden=True)
+        set_params_visibility([Param.MEMBERSHIP_ID], hidden=True)
+        set_params_visibility(TRANSFER_CUSTOMER_PARAMETERS, hidden=False)
+
+    return updated_order
```
svazquezco, line 115: I feel this logic a bit complicated and I'm not a big fan of the use of nonlocal except if there is a really good reason to use it. I believe we can use a map and the call the set_parameter_hidden and set_parameter_visible. Somethin like:
```
    parameters_map = {
        "new": {
            "visible": PARAM_NEW_CUSTOMER_PARAMETERS,
            "hidden": TRANSFER_CUSTOMER_PARAMETERS + (Param.MEMBERSHIP_ID,),
        },
        "Migrate": {
            "visible": [Param.MEMBERSHIP_ID],
            "hidden": PARAM_NEW_CUSTOMER_PARAMETERS + TRANSFER_CUSTOMER_PARAMETERS,
        
        },
        ...
    }
   update_order = set_parameter_visible(update_order, param_map[agreement_value] # I assume all the values have both visible and hidden params ,otherwise we can use an if conditional before with a get
   update_order = set_parameter_hidden(update_order, param_map[agreement_value] 

    return updated_order
```

adobe_vipm/flows/utils/validation.py
```diff
@@ -1,4 +1,7 @@
-from adobe_vipm.flows.constants import PARAM_REQUIRED_CUSTOMER_ORDER, Param
+from adobe_vipm.flows.constants import (
+    PARAM_REQUIRED_CUSTOMER_ORDER,
+    Param,
+)
```
svazquezco, line 4: I'm a bit confused with this change and the one on [line 63](https://github.com/softwareone-platform/swo-adobe-vipm-extension/pull/470/files#diff-0c923e4cd906b020dd1f11ef80237d9ab9e15b4e332ebbf0b4fcf71ff66147a7L63). Are they a Ruff change?

PR #476, opened 2025-07-25
MPT-11084 Process missing subscription with currency different than the main one

jentyk:


tests/flows/test_sync.py
```diff
@@ -2340,9 +2352,199 @@ def test_get_subscriptions_for_update_not_end_sale(
...
+                ],
+                "name": "Subscription for {agreement['product']['name']}",
+                "startDate": "2019-05-20T22:49:55Z",
+                "externalIds": {"vendor": "subscriptionId2"},
+                "product": {"id": "PRD-1111-1111"},
+                "autoRenew": True,
+            },
+        ),
+    ]
+
```
robcsegal, line 2505: Might be good to use a fixture factory for this. So much of these two is the same. A fixture factory could be used here.

adobe_vipm/flows/sync.py
```diff
@@ -43,11 +45,118 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+def _add_missing_subscriptions(
+    mpt_client: MPTClient,
+    adobe_client: AdobeClient,
+    customer: dict,
+    agreement: dict,
+    subscriptions_for_update: tuple[str],
```
d3rky, line 59: set?

adobe_vipm/flows/sync.py
```diff
@@ -43,11 +45,118 @@
...
+) -> None:
+    logger.info(f"Checking missing subscriptions for {agreement["id"]=}")
+    adobe_subscriptions = adobe_client.get_subscriptions(
```
d3rky, line 62: I think the problem with this function is that it returns all the subscriptions for customer, including all the deployment subscriptions that belongs to another deployment agreements.

d3rky (changes requested):

tests/flows/test_sync.py
jentyk: Are you referring to the expected mock calls? If yes, I don't think this is really the purpose of the fixtures. This would make reading the test more difficult if we start generating expected outputs.

adobe_vipm/flows/sync.py
```diff
@@ -160,7 +176,7 @@ def sync_agreement_prices(
         adobe_client,
         customer,
         agreement,
-        subscriptions_for_update=tuple(s[1]["subscriptionId"] for s in subscriptions_for_update),
+        subscriptions_for_update={s[1]["subscriptionId"] for s in subscriptions_for_update},
```
d3rky, line 179: I'm getting lost on `_get_subscriptions_for_update` cause it doesn't get subscriptions, it also terminates and updates it. So the naming/function/flow is really messy here :-( And also, I didn't get how do you solve my previous comment about deployments and that customer subscriptions contains subscription for all the deployments, not just for particular main agreement or deployment agreement

d3rky (changes requested):

adobe_vipm/flows/sync.py
jentyk: For now deployments are not being included. This is ticket no https://softwareone.atlassian.net/browse/MPT-11081

d3rky (approved):

ruben-sebrango (approved):

PR #489, opened 2025-08-01
MPT-11081 add missing subscriptions also for deployment agreements

jentyk:


alephsur (approved):

adobe_vipm/flows/sync.py
```diff
@@ -58,10 +58,15 @@ def _add_missing_subscriptions(
     agreement: dict,
     subscriptions_for_update: set[str],
 ) -> None:
-    logger.info(f"Checking missing subscriptions for {agreement["id"]=}")
-    adobe_subscriptions = exclude_subscriptions_with_deployment_id(
-        adobe_client.get_subscriptions(agreement["authorization"]["id"], customer["customerId"])
-    )["items"]
+    deployment_id = get_deployment_id(agreement) or ""
```
d3rky, line 61: `get_deployment_id` returns str or None. Why do you need an empty string here?

adobe_vipm/flows/sync.py
```diff
@@ -58,10 +58,15 @@ def _add_missing_subscriptions(
...
+    logger.info(f"Checking missing subscriptions for {agreement["id"]=}, {deployment_id=}")
```
d3rky, line 62: In order to make my life easier with new rules, could you stop using f-strings in log statements

adobe_vipm/flows/sync.py
```diff
@@ -58,10 +58,15 @@ def _add_missing_subscriptions(
...
+    adobe_subscriptions = [
+        i
```
d3rky, line 64: Pls, put more reasonable name for this variable. More than 1 symbol :-)

d3rky (changes requested):

adobe_vipm/flows/sync.py
jentyk: I need to compare this to `deploymentId` which is an empty string AFAIK. Just in case I make an empty string anyway.

d3rky (approved):

PR #491, opened 2025-08-04
MPT-11081 Lost Subscriptions in Adobe Account (in Deployment)

jentyk:


adobe_vipm/flows/sync.py
```diff
@@ -441,10 +442,12 @@ def _get_subscriptions_for_update(
         subscription = get_agreement_subscription(mpt_client, sub["id"])
         adobe_subscription_id = subscription["externalIds"]["vendor"]
 
-        adobe_subscription = find_first(
-            lambda x, subscr_id=adobe_subscription_id: x.get("subscriptionId", "") == subscr_id,
-            customer_subscriptions,
-        )
+        def is_searched_subscription(  # noqa: WPS430
```
d3rky, line 445: well, basically the idea is not remove `noqa`'s instead of adding new one :-)

adobe_vipm/flows/sync.py
```diff
@@ -862,6 +865,46 @@ def _update_last_sync_date(mpt_client: MPTClient, agreement: dict) -> None:
...
+        get_fulfillment_parameter(subscription, Param.ADOBE_SKU)["value"]
+        for agreement in deployment_agreements
+        for subscription in agreement["subscriptions"]
+    }
+    adobe_subscription_ids = {
+        subscription["subscriptionId"] for subscription in customer_subscriptions
+    }
+    orphaned_subscription_ids = adobe_subscription_ids - mpt_subscription_ids
+
+    def _is_orphaned_subscription(  # noqa: WPS430
```
d3rky, line 886: same here, let's try to remove `noqa`s

adobe_vipm/flows/utils/notification.py
```diff
@@ -18,7 +18,9 @@ def get_notifications_recipient(order: dict) -> str | None:
         Either contact parameter or buyer contact emails.
     """
     contact = get_ordering_parameter(order, Param.CONTACT.value).get("value", {}) or {}
-    buyer =order["agreement"]["buyer"].get("contact", {}) or {}return (contact or buyer).get("email")
+    buyer = order["agreement"]["buyer"].get("contact", {}) or {}
```
d3rky, line 21: Thank you for fixing it, my mistake :-(

pyproject.toml
```diff
@@ -184,6 +184,7 @@ pydocstyle.convention = "google"
   "S404", # subprocess calls are for tests
   "S603", # do not require `shell=True`
   "S607", # partial executable paths
+  "PLC2701", # Access to a protected member
```
d3rky, line 187: That's I would leave as is and refactor it later. Could you please revert changes of this commit?

d3rky (changes requested):

adobe_vipm/flows/sync.py
```diff
@@ -857,25 +922,35 @@ def sync_deployments_prices(
...
     deployment_agreements = get_agreements_by_customer_deployments(
         mpt_client,
-        Param.DEPLOYMENT_ID,
+        Param.DEPLOYMENT_ID.value,
         [deployment["deploymentId"] for deployment in customer_deployments],
     )
 
+    _process_orphaned_deployment_subscriptions(
+        adobe_client,
+        main_agreement["authorization"]["id"],
```
jentyk, line 940: @alephsur can you please confirm it is OK to pass here the`main_agreement["authorization"]["id"]`? Is the authorisation id the same for all subscriptions of a specific customer in a context of his all deployments and main agreement? # @d3rky

adobe_vipm/flows/sync.py
```diff
@@ -862,6 +868,49 @@ def _update_last_sync_date(mpt_client: MPTClient, agreement: dict) -> None:
...
+def _process_orphaned_deployment_subscriptions(
+    adobe_client: AdobeClient,
+    authorization_id,
+    customer_id,
+    deployment_agreements: list[dict],
+    customer_subscriptions: Sequence[dict],
+) -> None:
+    logger.info("Looking for orphaned deployment subscriptions in Adobe.")
+    mpt_subscription_ids = {
+        get_fulfillment_parameter(subscription, Param.ADOBE_SKU)["value"]
```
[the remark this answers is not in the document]
jentyk: It can, but in this case it means that we have corrupted data in our DB. Also, it would erroneously flag that subscription as not in the system, which would be worse. It is better if we get an exception logged to Teams for investigation in such a bizarre case.

adobe_vipm/flows/benefits.py
```diff
@@ -29,19 +29,19 @@
 def _build_3yc_parameters(request_info, commitment_info, is_recommitment):
     """Build parameters for 3YC commitment request."""
     status_param_ext_id = (
-        Param.THREE_YC_COMMITMENT_REQUEST_STATUS
+        Param.THREE_YC_COMMITMENT_REQUEST_STATUS.value
```
alephsur, line 32: Why do you need to use .value? Param is an StrEnum

tests/flows/test_sync.py
```diff
@@ -1319,141 +1105,65 @@ def test_sync_global_customer_update_not_required(
...
+        another_deployment_subscription,
+    ]
+    mock_get_prices_for_skus.side_effect = [
+        {"65304578CA01A12": 1234.55, "77777777CA01A12": 20.22},
+        {"65304578CA01A12": 1234.55, "77777777CA01A12": 20.22},
+        {"65304578CA01A12": 1234.55, "77777777CA01A12": 20.22},
+        {"65304578CA01A12": 1234.55, "77777777CA01A12": 20.22},
+        {"65304578CA01A12": 1234.55, "77777777CA01A12": 20.22},
+        {"65304578CA01A12": 1234.55, "77777777CA01A12": 20.22},
+    ]
```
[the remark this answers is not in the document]
jentyk: As I remember, I did it this way to fail the test if the number of calls is higher than the number of elements. However, if it was lower, it would not fail, so the check is not so consistent anyway, and it will be more readable with one liner, so I changed it.

alephsur (approved):

albertsola (approved):

PR #500, opened 2025-08-11
MPT-12504 Adobe CustomerDiscountsNotFoundError

alephsur:


adobe_vipm/flows/sync.py
```diff
@@ -767,6 +767,9 @@ def sync_agreement(  # noqa: C901
 
         try:
             customer = adobe_client.get_customer(agreement["authorization"]["id"], customer_id)
+            subscriptions = adobe_client.get_subscriptions(
```
jentyk, line 770: I would move it outside of `try`. This specific `try`has one single purpose and adding it here make things obscure. Also, AFAIK adobe client methods are wrapped in error handling anyway.

adobe_vipm/flows/sync.py
```diff
@@ -783,6 +783,15 @@ def sync_agreement(  # noqa: C901
                 return
             raise
 
+        subscriptions = adobe_client.get_subscriptions(
+            agreement["authorization"]["id"],
+            customer_id)
+
+        if not subscriptions.get("items", []):
```
jentyk, line 790: This snipped here logs "Skipping price sync", but in reality it also skips `_update_agreement` from line 805 and processing deployments, is it intentional? If it should only skip the price sync, it should be moved to the conditional in line 800 which specifically skips prices. If the whole rest of this function should be skipped, then it would be better to move it to line 756 and avoid what is being done below as this would be useless.

jentyk (approved):

robcsegal (approved):

PR #502, opened 2025-08-12
MPT-12643 Adobe - KeyError in 3YC request for an agreement

alephsur:


adobe_vipm/flows/benefits.py
```diff
@@ -80,12 +80,12 @@ def check_3yc_commitment_request(mpt_client, is_recommitment=False):
...
                     {
                         "externalId": status_param_ext_id,
-                        "value": request_info["status"],
+                        "value": request_info.get("status", None),
                     },
                 ]
             }
             logger.info(
-                f"3YC request for agreement {agreement['id']} is {request_info['status']}"
+                f"3YC request for agreement {agreement['id']} is {request_info.get("status", None)}"
```
jentyk, line 88: None is default for `get`, so you can skip it to make it simpler

PR #503, opened 2025-08-13
MPT-12679 Adobe - AdobeProduct with vendor external id not found

alephsur:


jentyk:
General comment, I think we would need some tests to cove the change. BTW it seems something wrong that no tests failed after implementing this change.

jentyk (approved):

albertsola (approved):

PR #515, opened 2025-08-27
MPT-12875 filter out one-time items when adding missing subscriptions

jentyk:


adobe_vipm/flows/sync.py
```diff
@@ -66,16 +67,15 @@ def _add_missing_subscriptions(
...
-        if line_item.get("deploymentId", "") == deployment_id
+        a_s for a_s in adobe_subscriptions if a_s.get("deploymentId", "") == deployment_id
     ]
-
+    one_time_skus = get_one_time_skus(mpt_client, agreement["product"]["id"])
     missing_subscriptions = tuple(
         subc
         for subc in deployment_subscriptions
-        if subc["subscriptionId"] not in subscriptions_for_update
+        if subc["subscriptionId"] not in subscription_for_update_ids
```
d3rky, line 76: `subsc`

adobe_vipm/flows/sync.py
```diff
@@ -1118,3 +1120,20 @@ def sync_all_agreements(mpt_client: MPTClient, *, dry_run: bool) -> None:
     agreements = get_all_agreements(mpt_client)
     for agreement in agreements:
         sync_agreement(mpt_client, agreement, dry_run=dry_run, sync_prices=False)
+
+
+def get_one_time_skus(mpt_client: MPTClient, product_id: str) -> tuple[str]:
```
d3rky, line 1125: I think it can be quite costly to fetch all the items to the memory each time. Better to pass SKUs that you already have in the list of subscriptions and fetch which ones are onetimers

d3rky (changes requested):

adobe_vipm/flows/sync.py
jentyk: https://github.com/softwareone-platform/mpt-extension-sdk/pull/63

jentyk: Done but requires SDK change (above) approval

jentyk: SDK v update

d3rky (approved):

PR #518, opened 2025-08-27
MPT-12387: add CreateOrUpdateAssets step and add it to the fullfill change…

svazquezco:
… order flow

### DIFF
--- a/adobe_vipm/flows/fulfillment/change.py
+++ b/adobe_vipm/flows/fulfillment/change.py
@@ -23,6 +23,7 @@ from adobe_vipm.flows.constants import (
 from adobe_vipm.flows.context import Context
 from adobe_vipm.flows.fulfillment.shared import (
     CompleteOrder,
+    CreateOrUpdateAssets,
     CreateOrUpdateSubscriptions,
     GetPreviewOrder,
     GetReturnOrders,
@@ -342,6 +343,7 @@ def fulfill_change_order(client, order):
         UpdateRenewalQuantities(),
         SubmitReturnOrders(),
         UpdateRenewalQuantitiesDownsizes(),
+        CreateOrUpdateAssets(),
         CreateOrUpdateSubscriptions(),
         UpdatePrices(),
         CompleteOrder(TEMPLATE_NAME_CHANGE),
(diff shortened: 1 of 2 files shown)


adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -918,6 +920,83 @@ def __call__(self, client, context, next_step):  # noqa: C901
...
+        for line in filter(
+            lambda x: get_partial_sku(x["offerId"]) in one_time_skus,
+            context.adobe_new_order["lineItems"],
+        ):
+            order_line = get_order_line_by_sku(context.order, line["offerId"])
+            adobe_subscription = adobe_client.get_subscription(
+                context.authorization_id, context.adobe_customer_id, line["subscriptionId"]
+            )
+
+            if adobe_subscription["status"] != AdobeStatus.PROCESSED:
```
jentyk, line 943: It is a good idea to always use `.value` if Enum is used

tests/conftest.py
```diff
@@ -1172,7 +1234,8 @@ def _order(
             else fulfillment_parameters
         )
 
-        lines = lines_factory(deployment_id=deployment_id) if lines is None else lines
+        asset = assets[0] if assets is not None else None
```
jentyk, line 1237: Why do we pass a sequence of assets and then gets the first one instead of passing single asset?

adobe_vipm/flows/fulfillment/shared.py
d3rky: You only need `.value` in fixtures when you assign value

adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -918,6 +920,83 @@ def __call__(self, client, context, next_step):  # noqa: C901
...
+            "name": f"Asset for {order_line['item']['name']}",
+            "parameters": {
+                "fulfillment": [
+                    {
+                        "externalId": Param.ADOBE_SKU.value,
+                        "value": line["offerId"],
+                    },
+                    {
+                        "externalId": Param.CURRENT_QUANTITY.value,
+                        "value": str(adobe_subscription[Param.CURRENT_QUANTITY.value]),
```
d3rky, line 981: Same here, you don't need `value`

```
>>> from enum import StrEnum
>>> class Enum(StrEnum):
...     A = "a"
...     B = "b"
...
>>> d = {"a": 1, "b": 2}
>>> d[Enum.A]
1
>>>
```

d3rky (approved):

PR #541, opened 2025-09-16
MPT-13411 Adobe - AdobeAPIError exceptions expired offer ids

alephsur:


ruben-sebrango (approved):

adobe_vipm/airtable/models.py
```diff
@@ -462,6 +462,35 @@ def get_prices_for_skus(product_id: str, currency: str, skus: list[str]) -> dict
     return {item.sku: item.unit_pp for item in items}
 
 
+def get_current_prices_availability_for_skus(
+    product_id: str, currency: str, skus: list[str]
+) -> dict:
```
d3rky, line 467: Invalid return type - should be list

adobe_vipm/airtable/models.py
```diff
@@ -462,6 +462,35 @@ def get_prices_for_skus(product_id: str, currency: str, skus: list[str]) -> dict
...
+
+    Args:
+        product_id: The ID of the product used to determine the AirTable base.
+        currency: The currency for which the price must be retrieved.
+        skus: List of SKUs which purchase prices must be retrieved.
+
+    Returns:
+        dict: A dictionary with SKU, purchase price items.
+    """
+    pricelist_model = get_pricelist_model(AirTableBaseInfo.for_pricing(product_id))
```
d3rky, line 481: Don't we have same function already? I have feeling that there is a function that does 2 things --> return current or 3yc prices. Maybe we can modify it and reuse first part here? It looks the same

adobe_vipm/flows/helpers.py
```diff
@@ -705,3 +731,47 @@ def __call__(self, mpt_client, context, next_step):
...
+class ValidateSkuAvailability(Step):
+    """Validate the SKU availability."""
+
+    def __init__(self, *, is_validation: bool) -> None:
+        self.is_validation = is_validation
+
+    def __call__(self, mpt_client, context, next_step):
+        """Validate the SKU availability."""
+        commitment = get_3yc_commitment(context.adobe_customer)
+        if commitment:
```
d3rky, line 745: Please put a comment here why you have here +1 year comparison. Is what Stu commented in the ticket.

adobe_vipm/flows/helpers.py
```diff
@@ -705,3 +731,47 @@ def __call__(self, mpt_client, context, next_step):
...
+            context.product_id, context.currency, adobe_skus
+        )
+        missing_skus = [sku for sku in adobe_base_skus if sku not in sku_prices]
+        if missing_skus:
+            logger.warning(
+                "SKU availability validation failed. Missing SKUs: %s. Available SKUs: %s",
+                missing_skus,
+                sku_prices,
+            )
+            context.validation_succeeded = False
```
d3rky, line 767: Didn't get why do you need it here...

setup.cfg
```diff
@@ -71,7 +71,7 @@ per-file-ignores =
   adobe_vipm/flows/context.py: WPS221 WPS237
   adobe_vipm/flows/errors.py: WPS110 WPS111 WPS202 WPS328 WPS430 WPS505
   adobe_vipm/flows/global_customer.py: WPS110 WPS111 WPS202 WPS210 WPS211 WPS212 WPS213 WPS229 WPS231 WPS232 WPS235 WPS237 WPS504
-  adobe_vipm/flows/helpers.py: WPS100 WPS110 WPS114 WPS118 WPS204 WPS210 WPS212 WPS213 WPS214 WPS219 WPS221 WPS231 WPS235 WPS338 WPS414 WPS504
+  adobe_vipm/flows/helpers.py: WPS100 WPS110 WPS114 WPS118 WPS202 WPS204 WPS210 WPS212 WPS213 WPS214 WPS219 WPS221 WPS231 WPS235 WPS338 WPS414 WPS432 WPS504
```
d3rky, line 74: Could you pls try to avoid adding new exceptions to the list? The idea is to make it smaller, but not bigger

d3rky (changes requested):

adobe_vipm/airtable/models.py
alephsur: I implemented this method only for the availability of the product, to get the list of products that exists in airtable without the prices.

setup.cfg
alephsur: The exceptions are included because of this two errors:
  0:1      WPS202 Found too many module members: 9 > 7

  747:92   WPS432 Found magic number: 365
  if commitment_end_date > dt.datetime.now(tz=dt.UTC).date() + dt.timedelta(days=365):

d3rky: Could we at least solve magic number? It is easy to do :-)

PR #552, opened 2025-09-22
MPT-12422 Improvements to Order Ordering and Fulfillment Parameters P…

alephsur:
…opulation

### DIFF
--- a/adobe_vipm/flows/fulfillment/transfer.py
+++ b/adobe_vipm/flows/fulfillment/transfer.py
@@ -259,6 +259,14 @@ def _fulfill_transfer_migrated(  # noqa: C901
         switch_order_to_failed(mpt_client, order, ERR_UPDATING_TRANSFER_ITEMS.to_dict())
         return
 
+    customer = adobe_client.get_customer(authorization_id, transfer.customer_id)
+    order = save_adobe_order_id_and_customer_data(
+        mpt_client,
+        order,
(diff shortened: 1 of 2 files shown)


adobe_vipm/flows/fulfillment/transfer.py
```diff
@@ -259,6 +259,14 @@ def _fulfill_transfer_migrated(  # noqa: C901
         switch_order_to_failed(mpt_client, order, ERR_UPDATING_TRANSFER_ITEMS.to_dict())
         return
 
+    customer = adobe_client.get_customer(authorization_id, transfer.customer_id)
+    order = save_adobe_order_id_and_customer_data(
+        mpt_client,
+        order,
+        transfer.transfer_id,
+        customer,
+    )
```
svazquezco, line 268: I'm wondering if we actually need this information to create a new order. I mean, why are we saving this data here instead of before -> [](https://github.com/softwareone-platform/swo-adobe-vipm-extension/blob/main/adobe_vipm/flows/fulfillment/transfer.py#L498-L504)

alephsur: This is to store the customer information before the creation of subscriptions. If there is any problem managing the rest elements of the agreements the customer information is not stored in the order.

ruben-sebrango (approved):

PR #575, opened 2025-10-06
MPT-14210 Sync expired subscription still displays Expiring templated…

alephsur:
… instead of Expired


adobe_vipm/flows/sync.py
```diff
@@ -577,6 +578,19 @@ def _get_subscriptions_for_update(
 
         if adobe_subscription["status"] == AdobeStatus.SUBSCRIPTION_TERMINATED:
             logger.info("Processing terminated Adobe subscription %s.", adobe_subscription_id)
+            template = mpt.get_template_by_name(
```
jentyk, line 581: I think it would be better to move it after `terminate_subscription` as this operation may fail, but you already changed the status to expired.

tests/flows/test_sync.py
```diff
@@ -1699,6 +1699,11 @@ def test_sync_agreement_prices_with_missing_prices(
     mock_adobe_client.get_subscriptions.assert_called_once_with("AUT-1234-5678", "a-client-id")
 
     assert mock_update_agreement_subscription.mock_calls == [
+        mocker.call(
+            mock_mpt_client,
+            terminated_mpt_subscription["id"],
+            template={"id": "TPL-1234", "name": "Renewing"},
```
jentyk, line 1705: Why the template is renewing, while the changes you are making in `sync.py` are `expiring`?

adobe_vipm/flows/sync.py
d3rky: I think there is no way to change the template on terminated subscription :-( That's why it is here. Am I correct @alephsur ?

alephsur: The subscription can be updated only in some cases. If the subscription is the only one in the agreement once is terminated the agreement is terminated also and the subscription cannot be updated.

tests/flows/test_sync.py
```diff
@@ -1882,11 +1890,52 @@ def test_get_subscriptions_for_update_terminated(
...
+    )
+    mock_terminate_subscription.assert_called_once_with(
+        mock_mpt_client,
+        mock_get_agreement_subscription.return_value["id"],
+        "Adobe subscription status 1004.",
+    )
+    mock_update_agreement_subscription.assert_called_once_with(
+        mock_mpt_client,
+        mock_get_agreement_subscription.return_value["id"],
+        template={"id": "TPL-1234", "name": "Renewing"},
```
jentyk, line 1916: Should this also be Expired?

jentyk (approved):

svazquezco (approved):

d3rky (approved):

PR #813, opened 2026-04-22
MPT-18859 Implement late renewals in change orders

alephsur:
Closes [MPT-18859](https://softwareone.atlassian.net/browse/MPT-18859)

- Implement late/manual renewals in change orders with two new pipeline steps:
  - CheckManualRenewalSubscriptions: detect subscriptions with MANUAL_RENEWAL allowed action, split lines into manual-renewal vs excess purchase lines, and persist late-renewal info.
  - SubmitRenewalOrders: preview and submit idempotent renewal orders (externalReferenceId = "{order_id}_RENEWAL"), reuse existing processed renewal orders when present, and store Adobe renewal orders in context.
- Integrate renewal steps into fulfillment pipelines for change, reseller-transfer/purchase, and transfer so renewals run before new-order submission.
- Adobe client additions:
  - OrderClientMixin.create_preview_renewal_for_subscription(...)
  - OrderClientMixin.create_renewal_order(...) (computes deterministic SHA-256 x-correlation-id)
- Context & persistence:
  - Add Context.manual_renewal_lines and Context.adobe_renewal_orders.
  - Add Param.LATE_RENEWALS_INFO = "lateRenewalsInfo" and migration to provision the ordering parameter for configured products.
- Line-item/subscription handling:
  - CreateOrUpdateSubscriptions now merges line items from adobe_new_order and adobe_renewal_orders, consolidating by subscriptionId and summing quantities.
  - SubscriptionClientMixin.get_subscriptions_for_offers treats subscriptions with MANUAL_RENEWAL_ACTION in allowedActions as active.
- Validation/flow tweaks:
  - ValidateDownsizes coterm-window logic tightened to only short-circuit when coterm-window applies and downsize_lines is non-empty.
  - CreateOrUpdateSubscriptions gating now proceeds when either adobe_new_order or manual_renewal_lines is present.
- Errors, constants & tests:
  - Add MANUAL_RENEWAL_ACTION constant and error codes VIPM0043 / VIPM0044 for preview/order failures.
  - Extensive tests added/updated covering detection, splitting, preview/submission, idempotency, failure paths, and pipeline ordering.
- Misc:
  - Migration added to create lateRenewalsInfo parameter.
  - Small flake8 per-file-ignore update and test fixture additions for the new ordering parameter.

[MPT-18859]: https://softwareone.atlassian.net/browse/MPT-18859?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -1368,3 +1396,295 @@ def __call__(self, client, context, next_step):
...
+        preview = self._validate_preview(adobe_client, client, context, ext_ref)
+        if preview is None:
+            return None
+
+        order = self._submit_order(adobe_client, client, context, ext_ref, preview)
+        if order is None:
+            return None
+
+        context.order = set_adobe_order_ids_created_parameter(context, [order["orderId"]])
+        return order
```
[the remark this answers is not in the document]
alephsur: The code only execute the order an store the parameters if it is not found in the external system. So the execution is correct in this case because if the order exists in the external system the existing order is returned.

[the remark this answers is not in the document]
alephsur: Yes, you are right

adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -1368,3 +1396,305 @@ def __call__(self, client, context, next_step):
...
+            excess_qty,
+        )
+
+        context.manual_renewal_lines[sku] = {
+            "line": line,
+            "adobe_subscription_id": adobe_sub_id,
+            "offer_id": adobe_subscription["offerId"],
+            "renewal_qty": renewal_qty,
+            "excess_qty": excess_qty,
+        }
```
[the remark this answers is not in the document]
alephsur: I have implemented the changes to recover the subscriptions using the deploymentId.

adobe_vipm/adobe/mixins/order.py
```diff
@@ -249,6 +249,86 @@ def create_preview_renewal(
         response.raise_for_status()
         return response.json()
 
+    @wrap_http_error
+    def create_preview_renewal_for_subscription(
```
d3rky: Hey, David! And could you as a separate commit refactor this mixins, I see there are couple of the WPS exception in pyproject toml file so I think we should try to remove one or two - step by step now

adobe_vipm/adobe/mixins/order.py
```diff
@@ -249,6 +249,86 @@ def create_preview_renewal(
...
+        authorization = self._config.get_authorization(authorization_id)
+        payload = {
+            "externalReferenceId": external_reference_id,
+            "currencyCode": authorization.currency,
+            "orderType": adobe_constants.ORDER_TYPE_RENEWAL,
+            "lineItems": line_items,
+        }
+        correlation_id = sha256(json.dumps(payload).encode()).hexdigest()
+        headers = self._get_headers(authorization, correlation_id=correlation_id)
+        response = requests.post(
```
d3rky, line 285: Looks like it is the same code as the previous function, the only difference is in the type of the order that you create. Could you pls review it? Maybe there are more similar functions

adobe_vipm/adobe/mixins/subscription.py
```diff
@@ -114,7 +114,10 @@ def get_subscriptions_for_offers(
             authorization_id, customer_id, deployment_id
         )["items"]
         active_subscriptions = filter(
-            lambda sub: sub["status"] == AdobeStatus.PROCESSED,
+            lambda sub: (
+                sub["status"] == AdobeStatus.PROCESSED
+                or MANUAL_RENEWAL_ACTION in sub.get("allowedActions", [])
```
d3rky: I don't think the we should mix active subscriptions and ones that we can apply manual renewal here. What scares me is that this function we use in quite a lot of functions, like agreement sync. And now we are treating subscription with manual renewal as an active one. But that's not like that :-( can we separate two flows here?

adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -1038,17 +1045,20 @@ class CreateOrUpdateSubscriptions(Step):
 
     def __call__(self, client, context, next_step):
         """Create or update subscriptions in MPT based on Adobe Subscriptions."""
-        if not context.adobe_new_order:
+        if not context.adobe_new_order and not context.manual_renewal_lines:
```
d3rky, line 1061: This `if` a little bit confuses me. So we are skipping the step if there is no any new adobe order exists. But it might me possible that it exists, but there are no manual renewal lines?

adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -1368,3 +1396,312 @@ def __call__(self, client, context, next_step):
...
+            {
+                "extLineItemNumber": line_item,
+                "offerId": info["offer_id"],
+                "quantity": info["renewal_qty"],
+                "subscriptionId": info["adobe_subscription_id"],
+            }
+            for line_item, info in enumerate(context.manual_renewal_lines.values(), start=1)
+        ]
+
+    def _validate_preview(self, adobe_client, client, context, ext_ref):
```
d3rky: I don't think that this step should be responsible for both things --> processing preview renewal and processing usual renewal order :-(

adobe_vipm/flows/validation/change.py
```diff
@@ -45,7 +45,7 @@ def _get_returnable_by_quantity_map(self, returnable_orders):
 
     def __call__(self, client, context, next_step):  # noqa: C901
         """Validates downsize items in order. Checks if it is possible to return them."""
-        if is_within_coterm_window(context.adobe_customer):
+        if is_within_coterm_window(context.adobe_customer) and context.downsize_lines:
```
d3rky: I didn't get why do we need this change here :-(

d3rky (changes requested):

adobe_vipm/adobe/mixins/subscription.py
alephsur: I have commited this change by mistake. I was making several test with the feature and I commited it.

adobe_vipm/flows/fulfillment/shared.py
alephsur: The context.adobe_new_order is the normal process and we skipped the process when that's not exists. But with the late_renewals we can have an order that is not creating a new order and we only have the context.manual_renewal_lines because the subscriptions that we are creating have the MANUAL_RENEWAL. So, in this case we are cannot skip the process because we are going to create the subscriptions for the items selected by the customer. 
That's why we are merging the lines stored in both to create the subscriptions with the correct quantities.

adobe_vipm/adobe/mixins/order.py
alephsur: I have removed the function create_preview_renewal_for_subscription to use the same function sending if the order is preview_renewal or create.

adobe_vipm/flows/fulfillment/shared.py
alephsur: I have splited the method to have twot diferent methods in the pipeline one for the preview and another one to create.

adobe_vipm/flows/validation/change.py
alephsur: I was making some testing a I commited this change by mistake. I have removed the code

d3rky (approved):

PR #839, opened 2026-06-01
MPT-18860 Discounts not found during sync process

jentyk:
Closes [MPT-18860](https://softwareone.atlassian.net/browse/MPT-18860)

- Add fallback for missing Adobe discounts: helpers now tolerate absent discount records and call sites default to a fake discount level ("T1" for consumables, "01" for licenses) when Adobe omits discounts.
- Recover instead of raising: discount-level lookup failures log a warning, invoke the notification helper, and continue using the fallback rather than throwing.
- Improved notifications: Teams/exception messages include the customer's discounts payload and state that a default level is being applied.
- Logging: added module-level logger in subscription utilities to surface discount-level warnings (includes customerId when available).
- Tests updated: sync and utils tests adjusted to expect notification calls and graceful recovery (notification receives full customer payload) instead of exceptions.

[MPT-18860]: https://softwareone.atlassian.net/browse/MPT-18860?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## ATTENTION :exclamation:

This is a workaround it defaults to a basic discount level in cases when Adobe issue described in related Jira ticket occurs. In such a case fake discount level will be assigned. It may happened in following calls:
- AssetSyncer._get_assets_for_update(self) (AssetSyncer in adobe_vipm.flows.sync.asset)
    AssetSyncer.sync(self, *, dry_run) (AssetSyncer in adobe_vipm.flows.sync.asset)
- AgreementSyncer._add_missing_subscriptions_and_assets(self) (2 usages) (AgreementSyncer in adobe_vipm.flows.sync.agreement)
    AgreementSyncer.sync(self, *, sync_prices) (AgreementSyncer in adobe_vipm.flows.sync.agreement)
- AgreementSyncer._get_subscriptions_for_update(self, agreement) (AgreementSyncer in adobe_vipm.flows.sync.agreement)
    AgreementSyncer.sync(self, *, sync_prices) (AgreementSyncer in adobe_vipm.flows.sync.agreement)
    AgreementSyncer._sync_deployment_agreements(self, adobe_deployments, deployment_agreements, *, sync_prices) (AgreementSyncer in adobe_vipm.flows.sync.agreement)
- process_adobe_subscriptions_from_agreement(mpt_client, adobe_client, adobe_subscriptions, adobe_customer, gc_agreement_id, agreement_deployment, buyer_id, product_id, authorization_id) (3 usages) (adobe_vipm.flows.global_customer)
    process_agreement_deployment(mpt_client, mpt_o_client, adobe_client, agreement_deployment, product_id) (adobe_vipm.flows.global_customer)

### DIFF
--- a/adobe_vipm/flows/utils/customer.py
+++ b/adobe_vipm/flows/utils/customer.py
@@ -120,9 +120,9 @@ def get_customer_licenses_discount_level(customer: dict) -> int:
         Adobe discount level for licensees.
     """
     licenses_discount = find_first(
-        lambda discount: discount["offerType"] == OfferType.LICENSE, customer["discounts"]
+        lambda discount: discount["offerType"] == OfferType.LICENSE, customer["discounts"], {}
     )
-    return licenses_discount["level"]
+    return licenses_discount.get("level")
 
 
 def get_customer_consumables_discount_level(customer: dict) -> int:
@@ -136,9 +136,9 @@ def get_customer_consumables_discount_level(customer: dict) -> int:
         Adobe discount level for consumables.
     """
     licenses_discount = find_first(
-        lambda discount: discount["offerType"] == OfferType.CONSUMABLES, customer["discounts"]
+        lambda discount: discount["offerType"] == OfferType.CONSUMABLES, customer["discounts"], {}
     )
-    return licenses_discount["level"]
+    return licenses_discount.get("level")
 
 
 def is_new_customer(source: dict) -> bool:
--- a/adobe_vipm/flows/utils/subscription.py
+++ b/adobe_vipm/flows/utils/subscription.py
@@ -1,4 +1,5 @@
 import datetime as dt
+import logging
 
 from mpt_extension_sdk.mpt_http.utils import find_first
 
@@ -14,6 +15,8 @@ from adobe_vipm.flows.utils.customer import (
     get_customer_licenses_discount_level,
 )
 from adobe_vipm.flows.utils.notification import notify_discount_level_error
+
+logger = logging.getLogger(__name__)
 
 
 def get_subscription_by_line_and_item_id(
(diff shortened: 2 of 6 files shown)


adobe_vipm/flows/utils/customer.py
```diff
@@ -120,9 +120,9 @@ def get_customer_licenses_discount_level(customer: dict) -> int:
         Adobe discount level for licensees.
     """
     licenses_discount = find_first(
-        lambda discount: discount["offerType"] == OfferType.LICENSE, customer["discounts"]
+        lambda discount: discount["offerType"] == OfferType.LICENSE, customer["discounts"], {}
     )
-    return licenses_discount["level"]
+    return licenses_discount.get("level")
```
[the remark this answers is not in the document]
jentyk: ATM missing `discounts` key in `customer` data structure would mean broken API and it should result in an exception.

adobe_vipm/flows/utils/subscription.py
```diff
@@ -143,15 +146,18 @@ def get_sku_with_discount_level(sku: str, customer: dict) -> str:
...
-    except Exception:
+    discount_level = (
+        get_customer_consumables_discount_level(customer)
+        if is_consumables_sku(sku)
+        else get_customer_licenses_discount_level(customer)
+    )
+    # FIXME: Workaround: Default to level 01 when customer has no discount level should be removed
+    #  after Adobe side fix the issue related to MPT-18860.
+    if not discount_level:
+        discount_level = "01"
```
alephsur: You are hardcoding the discount level with 01, but that is not valid for consumables. Consumables are T1,T2, etc.
Even in the 3YC is applied 01 is not valid. When you are hardcoding the value you can create a problem setting a wrong value. I think that we should break the process and not set a wrong value that we don't know. The issue is in Adobe side that when 3YC is applied on for licenses they are not returning the discounts for consumables.

alephsur (changes requested):

adobe_vipm/flows/utils/subscription.py
jentyk: @smeeks-swo ?

smeeks-swo: @jentyk this is true. '01' only applies to licenses, not consumables. But I still think we can have a win here:

update:

https://github.com/softwareone-platform/swo-adobe-vipm-extension/blob/03f2632a804da75568b082ab22f34f30ec9d4b23/adobe_vipm/flows/utils/customer.py#L112

to return "01" if the value is absent, and

and update:

https://github.com/softwareone-platform/swo-adobe-vipm-extension/blob/03f2632a804da75568b082ab22f34f30ec9d4b23/adobe_vipm/flows/utils/customer.py#L128

to return "T1" if the value is absent.

I know this is not ideal, but it is far, far better than agreement sync failing

@d3rky what do you think?

d3rky: I'm ok with it, but still we should notify in teams + warning in logs

jentyk: Done

adobe_vipm/flows/utils/subscription.py
```diff
@@ -143,15 +146,25 @@ def get_sku_with_discount_level(sku: str, customer: dict) -> str:
...
+    #  fail. Should be removed after Adobe side fix the issue.
+    if not discount_level:
+        discount_level = "T1" if is_consumable else "01"
+        logger.warning(
+            "Customer %s has no discount level, defaulting to level %s",
+            customer.get("customerId"),
+            discount_level,
         )
-    except Exception:
         notify_discount_level_error(sku, customer)
```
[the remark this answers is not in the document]
jentyk: done

alephsur (approved):

adobe_vipm/flows/utils/subscription.py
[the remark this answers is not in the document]
jentyk: tests passing now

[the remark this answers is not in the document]
jentyk: This is a temp workaround. No point of passing. Genral message is fine

PR #901, opened 2026-07-14
MPT-23189 use renewalQuantity instead of currentQuantity when creating subscriptions

smeeks-swo:
🤖 AI-generated PR — Please review carefully.

## What

`_create_mpt_subscription` set the initial MPT subscription line quantity from Adobe's `currentQuantity` field. The ongoing sync path, `SubscriptionSyncer._update_subscription`, correctly uses `autoRenewal.renewalQuantity` for the same field — that is the value MPT users are meant to see going forward.

This PR aligns subscription creation with the sync path by reading `adobe_subscription["autoRenewal"]["renewalQuantity"]` instead of `adobe_subscription["currentQuantity"]` when building the initial subscription's line quantity. Asset creation (`_create_mpt_asset`) is unaffected — one-time assets have no renewal concept and correctly keep using `currentQuantity`.

## Why

`currentQuantity` and `renewalQuantity` can genuinely differ (e.g. a customer has a pending seat change scheduled for the next renewal). Without this fix, a newly created subscription could show the wrong quantity until the next scheduled sync run corrected it.

## Testing

- Added `test_add_missing_subscriptions_uses_renewal_quantity_not_current_quantity`, which uses distinct current/renewal quantities (5 vs. 8) to pin the behavior. The existing creation tests default both values to the same number and would not have caught this regression.
- `make check-all` (ruff format, ruff check, flake8, uv lock check, full test suite — 896 tests) passes locally.

Jira: MPT-23189

Closes [MPT-23182](https://softwareone.atlassian.net/browse/MPT-23182)

- Updated MPT subscription line quantity creation to use Adobe `autoRenewal.renewalQuantity` (instead of `currentQuantity`) to align with renewal-quantity synchronization.
- Kept asset creation using `currentQuantity`.
- Added a regression test where current and renewal quantities differ (e.g., 7 vs 10) to verify the created agreement subscription payload uses renewal quantity for the subscription line while preserving `fulfillment.currentQuantity`.

[MPT-23182]: https://softwareone.atlassian.net/browse/MPT-23182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/flows/sync/test_agreement.py
+++ b/tests/flows/sync/test_agreement.py
@@ -3222,7 +3222,10 @@ def test_add_missing_subscriptions(
             subscription_id="1e5b9c974c4ea1bcabdb0fe697a2f1NA", offer_id="65322572CAT1A10"
         ),
         adobe_subscription_factory(
-            subscription_id="2e5b9c974c4ea1bcabdb0fe697a2f1NA", offer_id="65322572CAT1A13"
+            subscription_id="2e5b9c974c4ea1bcabdb0fe697a2f1NA",
+            offer_id="65322572CAT1A13",
+            current_quantity=7,
+            renewal_quantity=10,
         ),
         adobe_subscription_factory(
             subscription_id="ae5b9c974c4ea1bcabdb0fe697a2f1NA", offer_id="75322572CAT1A11"
@@ -3303,7 +3306,7 @@ def test_add_missing_subscriptions(
             "parameters": {
                 "fulfillment": [
                     {"externalId": "adobeSKU", "value": "65322572CAT1A13"},
-                    {"externalId": "currentQuantity", "value": "10"},
+                    {"externalId": "currentQuantity", "value": "7"},
                     {"externalId": "renewalQuantity", "value": "10"},
                     {"externalId": "renewalDate", "value": "2026-10-11"},
                 ]
--- a/adobe_vipm/flows/sync/agreement.py
+++ b/adobe_vipm/flows/sync/agreement.py
@@ -380,7 +380,7 @@ class AgreementSyncer:  # noqa: WPS214
             "seller": {"id": self._seller_id},
             "lines": [
                 {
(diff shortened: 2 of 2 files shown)


tests/flows/sync/test_agreement.py
```diff
@@ -3197,6 +3197,54 @@ def test_add_missing_subscriptions_recreates_when_mpt_subscription_is_expired(
     assert create_payload["externalIds"]["vendor"] == expired_external_id
 
 
+@freeze_time("2025-07-24")
+def test_add_missing_subscriptions_uses_renewal_quantity_not_current_quantity(
```
d3rky: Mmm... I'm a little bit surprised here that we have additional test here. I'm pretty sure that we already have a test that checks current quantity, but for some reason it is not failing after your changes. It only means that tests is written in a incorrect way.

That's why I would better to fix existing test instead of the adding new one

d3rky (changes requested):

tests/flows/sync/test_agreement.py
smeeks-swo: Good catch — fixed the existing `test_add_missing_subscriptions` instead of adding a new one. It was asserting on the created subscription's quantity but used a fixture where `current_quantity` and `renewal_quantity` were equal (both 10), so it couldn't distinguish which field was actually used. Gave that subscription `current_quantity=7, renewal_quantity=10` and updated the `currentQuantity` fulfillment-param assertion to `7` — the `quantity` line assertion stays `10`, now meaningfully pinning `renewalQuantity` as the source. Removed the redundant test I'd added. Verified by temporarily reverting the source fix locally and confirming this test fails, then restored it — all green (`make check-all`, 895 passed).

🤖 Generated by AI

d3rky (approved):

PR #931, opened 2026-07-29
MPT-23679 Switch order are failing when we use the same skus in diffe…

alephsur:
…rent switches

- Updated switch-order validation to allow existing SKUs across different switches.
- Preserved detection of duplicate lines within the same switch order.
- Added test coverage for both scenarios.

### DIFF
--- a/adobe_vipm/flows/fulfillment/shared.py
+++ b/adobe_vipm/flows/fulfillment/shared.py
@@ -1329,6 +1329,13 @@ class SyncAgreement(Step):
 class ValidateDuplicateLines(Step):
     """Validates if Adobe Order contains duplicated items, with the same sku."""
 
+    def __init__(self, *, is_switch=False):
+        # Switch (mid-term upgrade) orders may legitimately target an item that already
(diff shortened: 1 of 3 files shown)


adobe_vipm/flows/fulfillment/shared.py
```diff
@@ -1341,22 +1348,23 @@ def __call__(self, client, context, next_step):
...
-        ])
-        duplicates = [item for item, count in Counter(items).items() if count > 1]
-        if duplicates:
-            switch_order_to_failed(
-                client,
-                context.order,
-                ERR_EXISTING_ITEMS.to_dict(duplicates=",".join(duplicates)),
-            )
-            return
+        if not self.is_switch:
```
d3rky, line 1351: It looks like for me that it should be 2 separate validation steps - one that validates duplicated lines in order lines (first part of the `__call__` function and another steps that validates duplicated lines in the agreement subscriptions. And in this case switch pipeline is going to have both steps, and other pipelines only the first one - WDYT?

d3rky (changes requested):

